PR #11, opened 2018-04-06
Simplify model, get rid of Transactions.

pudo:
I think this would be more explicit and result in a clearer UI.


pudo:
I doubt it, too.

PR #24, opened 2018-09-12
allow for format as a year

uhhhuh:


pudo:
I don't think we're done here, this doesn't handle a case like ``%m.%Y``.

To give an even worse idea: what if we didn't do regex but instead checked the format string to determine it's granularity?

```python
if '%d' not in format:
    # chop off the day part of the stringified outcome
if '%m' not in format:
    # chop off the month part of the outcome.
```

Needs lots of unit testing :/

PR #29, opened 2018-12-17
Making the TaxRoll schema a bit more robust

Ueland:


pudo:
Since you're the only user of the `TaxRoll` I'll just add this in to not block you. 

I'm concerned that we still might be talking across each other a little bit here. To me, the schema of a `Sanction` or `Passport` seems like a good equivalent to the `TaxRoll`: it's essentially an add-on to a `Person`. You don't really want a `Passport` to show up in your search results (which is what making it a `Thing` does), but rather the person it is associated with. Then, when you open that person's profile, you will see a list of associated passports or sanction embargoes for that individual. Translated to tax declarations that would mean: there's a profile for `Person` X, and several associated `TaxRoll` entities for the years 2017, 2016 etc.

Ueland:
Thank you for a good explanation @pudo! I still have some to learn here. I totally agree with what you are saying and will do the nesessary changes and send in a new PR, hopefully today.

PR #37, opened 2019-02-25
CSV and Excel exporters

### LINKED ISSUE #30: Exporter for CSV and Excel
For a given stream of entities, let's implement an abstraction to: 

1. Serialise each entity into a table entry, with the table named after the entity schema.
2. Merge multi-valued properties in a semantically useful way. Default could just be `'; '.join(values)`
3. Sort the output columns so that featured properties appear first, alphabetically after that.
4. Also have convenience functions to write to CSV and OpenPyxl. 

Also, let's implement a CLI version of this in `util/followthemoney_util`.

sunu:
Fixes #30

### DIFF
--- a/followthemoney/export/csv.py
+++ b/followthemoney/export/csv.py
@@ -1,0 +1,18 @@
+import csv
+
+
+def write_entity(fh, entity):
+    fieldnames = [prop.label for prop in entity.schema.sorted_properties]
+    fieldnames = ['id', ] + fieldnames
+    writer = csv.DictWriter(fh, fieldnames=fieldnames)
+    prop_dict = {'id': entity.id}
+    for prop in entity.schema.sorted_properties:
+        prop_dict[prop.label] = prop.type.join(entity.get(prop))
+    writer.writerow(prop_dict)
+
+
+def write_headers(fh, schema):
+    fieldnames = [prop.label for prop in schema.sorted_properties]
+    fieldnames = ['id', ] + fieldnames
+    writer = csv.DictWriter(fh, fieldnames=fieldnames)
+    writer.writeheader()
--- a/followthemoney/export/csv.py
+++ b/followthemoney/export/csv.py
@@ -1,18 +1,21 @@
 import csv
 
 
-def write_entity(fh, entity):
+def write_entity(fh, entity, extra_fields=dict()):
     fieldnames = [prop.label for prop in entity.schema.sorted_properties]
-    fieldnames = ['id', ] + fieldnames
+    fieldnames = ['id'] + list(extra_fields.keys()) + fieldnames
     writer = csv.DictWriter(fh, fieldnames=fieldnames)
-    prop_dict = {'id': entity.id}
+    prop_dict = {
+        'id': entity.id,
+        **extra_fields
+    }
     for prop in entity.schema.sorted_properties:
         prop_dict[prop.label] = prop.type.join(entity.get(prop))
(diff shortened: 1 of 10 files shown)


pudo:
Very nice! Should we pass in an output directory as a parameter?

pudo:
Two comments: should we also emit the ID for each row? And should part of this move to ftm core so that it can be used from aleph?

sunu:
Yeah, adding ftm-util as an additional dependency for this only is probably not ideal. Where should this code ideally live inside core? `followthemoney/export/`?

pudo:
That sounds clear!

util/followthemoney_util/excel.py
```diff
@@ -0,0 +1,29 @@
...
+
+from followthemoney_util.cli import cli
+from followthemoney_util.util import read_object
+from followthemoney.export.excel import get_workbook, get_sheet, write_entity
+
+
+@cli.command('export-excel', help="Export to Excel")
+@click.option('--filename', default='export.xlsx',
+              help="name of the excel file")
+@click.option('--outdir', type=click.Path(exists=True), default='.',
```
pudo: I feel this might be redundant with `filename`, perhaps we can use validation on that (i.e. make it a `click.Path`) and remove `outdir`.

util/followthemoney_util/excel.py
```diff
@@ -0,0 +1,29 @@
...
+              help="output directory")
+def export_excel(filename, outdir):
+    stdin = click.get_text_stream('stdin')
+    workbook = get_workbook()
+    try:
+        while True:
+            entity = read_object(stdin)
+            if entity is None:
+                break
+            sheet = get_sheet(entity.schema, workbook)
```
pudo: We could move this into `write_entity`, but that's cosmetics.

util/followthemoney_util/csv.py
```diff
@@ -0,0 +1,37 @@
...
+
+from followthemoney_util.cli import cli
+from followthemoney_util.util import read_object
+from followthemoney.export.csv import write_entity, write_headers
+
+
+def _get_csv_handler(outdir, schema, handlers):
+    fh = handlers.get(schema.name)
+    if fh is None:
+        name = "{0}.csv".format(schema.plural)
```
pudo: This is really cool. Do we need to do something with the schema name to make it safe? I don't think so but it's worth thinking twice :)

followthemoney/export/csv.py
```diff
@@ -0,0 +1,21 @@
+import csv
+
+
+def write_entity(fh, entity, extra_fields=dict()):
+    fieldnames = [prop.label for prop in entity.schema.sorted_properties]
+    fieldnames = ['id'] + list(extra_fields.keys()) + fieldnames
```
pudo: I assume extra fields is stuff like the UI url in Aleph? Very cool. But is there a risk with the method-default dict on L4? Maybe let's make this None and then zero-guard it.

sunu: Yeah, the default needs to be `None`. I mistakenly thought making it `dict()` instead of `{}` will help.

util/followthemoney_util/csv.py
sunu: The name sometimes has space in it. Should we slugify it?

util/followthemoney_util/excel.py
sunu: Yeah, makes sense.

PR #54, opened 2019-08-07
Cast dates and numeric properties into an approximate float property

sunu:


pudo:
I'd split the responsibilities differently here: the `.num` exists only to solve a problem intrinsic to ElasticSearch, so I don't think we should introduce them in this library. I'd rather have the cast methods (`_cast_float`, `_cast_datetime`) here, and keep the rest in aleph, e.g. in `aleph.index.entities:format_proxy`.

followthemoney/types/date.py
```diff
@@ -95,3 +95,23 @@ def compare(self, left, right):
...
+        formats = [
+            "%Y-%m-%d %H:%M:%S",
+            "%Y-%m-%d %H:%M",
+            "%Y-%m-%d %H",
+            "%Y-%m-%d",
+            "%Y-%m",
+            "%Y",
+        ]
+        date = None
+        for fmt in formats:
```
pudo: One nitpick: this is gonna get run millions upon millions of times. I'm wondering it it would be worthwhile putting the formats into a class-level dict that's indexed by string length, i.e.: `{4: '%Y'}`. That way, rather than trying each format, the correct one could be inferred in O(1) and only one call to `strptime` is required.

followthemoney/types/number.py
```diff
@@ -7,3 +7,6 @@ class NumberType(PropertyType):
     label = _('Number')
     plural = _('Numbers')
     matchable = False
+
+    def _cast_num(self, value):
+        return float(value)
```
pudo: Should we regex away anything that isn't a `[0-9\-\.]` first?

PR #101, opened 2019-11-12
Another one bites the dust

dchaplinsky:
@pudo , I think you've forgot to close one handler.

### DIFF
--- a/followthemoney/export/neo4j.py
+++ b/followthemoney/export/neo4j.py
@@ -115,6 +115,7 @@ class Neo4JCSVExporter(CSVExporter, GraphExporter):
                     fp.write(cmd.format(file_name))
 
         self.links_handler.close()
+        self.nodes_handler.close()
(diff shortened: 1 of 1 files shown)


dchaplinsky:
First commit didn't completely fixed it because entities are written into `_links.csv` but not to `_nodes.csv`. I've made a fix for that but I need to discuss if it's correct thing to do. Maybe we don't need to write links that goes to rdf.type = entity?

pudo:
I'm pretty sure the exemption for entity nodes is correct since they *must* be in one of the schema-specific CSV files, unless the provided data is invalid. Do you get errors out of this?

dchaplinsky:
Yes, I do, because they are also mentioned in _links.csv with entity:
prefix and Neo4j cannot find them. I'll look deeply into it today.

pudo:
Discard in favour of #105

PR #107, opened 2019-11-14
Fixing entity: prefix for edges

dchaplinsky:


dchaplinsky:
@pudo, please pay attention to comment `# That potentially may lead to multiple edges with same id`

PR #115, opened 2019-12-12
Create All Individual Entity Type List

tendai-zw:
A list of all Individual Entities within the Follow the Money data model


pudo:
heya, thanks for your PR! I wonder if there is some way that this can be generated automatically, rather than writing it down in duplicate. For example, you can always browse the ontology here:

* http://150.146.207.114/lode/extract?url=https%3A%2F%2Falephdata.github.io%2Ffollowthemoney%2Fns%2Fftm.xml&owlapi=true&imported=true&lang=en 

And as a network diagram here:

* http://www.visualdataweb.de/webvowl/#iri=https://alephdata.github.io/followthemoney/ns/ftm.xml

tendai-zw:
> heya, thanks for your PR! I wonder if there is some way that this can be generated automatically, rather than writing it down in duplicate. For example, you can always browse the ontology here:
> 
> http://150.146.207.114/lode/extract?url=https%3A%2F%2Falephdata.github.io%2Ffollowthemoney%2Fns%2Fftm.xml&owlapi=true&imported=true&lang=en
> 
> And as a network diagram here:
> 
> http://www.visualdataweb.de/webvowl/#iri=https://alephdata.github.io/followthemoney/ns/ftm.xml

Thanks a lot,  It should be possible to generate automatically, perhaps when i get free time i can try to figure out something. 
I am working on an Aleph instance and felt maybe someone would need the complete list like me.  

Thanks for those links, very helpful. They are exactly what i needed.

pudo:
I've added that to the documentation here: 

https://docs.alephdata.org/developers/followthemoney#browsing-the-schema

But I might close the PR, if you don't mind, until we decide we need some other automatic mechanism.

PR #145, opened 2020-01-30
Maintain the order of property values

sunu:
VIS captions are based on the first value from the list of values. Without some fixed order, caption will change randomly in the UI.

Alternatively, instead of insertion order, we can maintain an order based on insertion count?


pudo:
Can we guarantee this survives through ElasticSearch? Otherwise I feel it’d be better not to make that promise :/

sunu:
Added a test to Aleph to check if it survives through ES https://github.com/alephdata/aleph/commit/f63a93f8045b42ef111de907245190baff44cf35. Passes with this change.

Issue #161, opened 2020-02-26
Add FtM schema for Calls, Messages, UserAccounts and Locations


sunu:
As part of https://github.com/alephdata/aleph/issues/1028 and going forward, we need these schema to model data into FtM land.

Here's the rough outline we discussed:
```
UserAccount <- Thing
name
email
number
service
password
owner [LegalEntity]
---

Call <- Interval
caller [LegalEntity]
callerNumber
receiver [LegalEntity] -- 1...n
receiverNumber
duration
(( recording [Audio] ))
---

Communication <- Interval
---

Message <- Communication
(i.e. SMS, InstantMessage)
sender [LegalEntity]
senderAccount [UserAccount]
receiver [LegalEntity] -- 1...n
receiverAccount [UserAccount]
Content
Attachment
Timestamp [date]
(( conversationID ))
(( sequenceNumber ))
inReplyTo [Message]
deleted
---

Location
(i.e. Journey, Location, BluetootDevice?, WirelessNetwork?)
```

pudo:
Merged

PR #162, opened 2020-02-26
Introduce Calls, Messages and UserAccounts

sunu:
I decided not to add a `Communication` schema because the property names sort of vary among Call, Message and Email. For example, for calls we have `caller` as an `LegalEntity`, for Messages it's `sender`, but for emails `sender` is a string right now. And if we change that we have to reprocess a lot of data which I assume is not ideal? Should we make `Communication` just a symbolic class for now? (without any properties of its own)

refs #161

### DIFF
--- a/followthemoney/schema/Message.yaml
+++ b/followthemoney/schema/Message.yaml
@@ -1,0 +1,71 @@
+Message:
+  extends:
+    - Interval
+    - Folder
+    - PlainText
+    - HyperText
+  label: Message
+  plural: Messages
+  matchable: false
+  generated: true
+  featured:
+    - subject
+    - threadTopic
+    - date
+    - sender
+    - recipients
+  caption:
+    - subject
+    - threadTopic
+    - title
+    - fileName
+  properties:
+    subject:
+      label: Subject
+      type: string
+    threadTopic:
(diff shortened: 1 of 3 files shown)


followthemoney/schema/Message.yaml
```diff
@@ -0,0 +1,71 @@
...
+      reverse:
+        name: messagesSent
+        label: "Messages sent"
+    senderAccount:
+      label: "Sender Account"
+      type: entity
+      range: UserAccount
+      reverse:
+        name: messagesSent
+        label: "Messages sent"
```
sunu: Don't know how to deal with this. I'm basically duplicating the same data here for `messagesSent` which is not ideal.

followthemoney/schema/UserAccount.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - username
+  - email
+  - service
+  properties:
+    owner:
+      label: "Owner"
+      type: entity
+      range: LegalEntity
+      reverse:
+        name: useraccounts
```
pudo: Camel case?

sunu:
@pudo Done!

PR #181, opened 2020-03-19
Make property expansion optional when adding a node

sunu:
graph

Suppose I have an expansion candidate like the graph in the image where `PersonA` and `PersonB` share a name and `PersonA` is director of `CompanyX` through `DirectorshipX`. And `PersonA` entity is the source of the expansion.

To construct the graph, we'll do:
```
graph.add(person_a)
graph.add(directorship_x)
graph.add(person_b)
```
Without any arguments graph doesn't know that we're interested in expanding property nodes of `PersonA` but not of `PersonB`. https://github.com/alephdata/followthemoney/blob/c9dc19150751f01d2cdddfc5f17273509a2d8dc6/followthemoney/graph.py#L137 will create nodes for the properties of `PersonB`. So we'll end up with extra layer of property expansion like in the image (the email node coming out of `PersonB` should not be created when expanding `PersonA`)

### DIFF
--- a/followthemoney/graph.py
+++ b/followthemoney/graph.py
@@ -127,25 +127,26 @@ class Graph(object):
         if edge.weight > 0:
             self.edges[edge.id] = edge
 
-    def _add_node(self, proxy):
+    def _add_node(self, proxy, expand_properties=True):
         """Derive a node and its value edges from the given proxy."""
         entity = Node(registry.entity, proxy.id, proxy=proxy)
         self.nodes[entity.id] = entity
         for prop, value in proxy.itervalues():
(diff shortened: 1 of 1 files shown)


pudo:
Is this still an issue if the return type of the API is entities instead of graph fragments?

sunu:
@pudo Actually we should revert this change. Because the API doesn't need it anymore after switching the return type from graph fragments to entities and also the change prevents the expansion of the common property link (`name` in the diagram above). So it's not very useful.

pudo:
@sunu No biggie. Wanna make a new PR?

PR #209, opened 2020-04-14
Add support for release, compileRelease

uhhhuh:
Include separate release update and aggregated compiled releases from ocds.


followthemoney/cli/ocds.py
```diff
@@ -137,8 +140,19 @@ def convert_release(release):
 def convert_record(record):
     published_date = clean_date(record.pop('publishedDate', None))
     publisher = record.pop('publisher', {}).get('name')
-    for release in record.get('releases', []):
-        for entity in convert_release(release):
+    if record.get('tag'):
+        for entity in convert_release(record):
```
pudo, line 144: Would it make sense to rename `convert_release` to a more generic term that covers both releases and records? `convert_item`?

followthemoney/cli/ocds.py
```diff
@@ -82,7 +85,7 @@ def convert_release(release):
 
     tender = release.pop('tender', {})
     contract = model.make_entity('Contract')
-    contract.make_id(release.pop('id', None))
+    contract.make_id(release.pop('ocid', None))
```
pudo, line 88: Should we play more of the "if you have X, use X, otherwise use Y" game here?

followthemoney/cli/ocds.py
```diff
@@ -10,6 +10,8 @@
 log = logging.getLogger(__name__)
 IDENTIFIERS = {
     'TRADE_REGISTER': 'registrationNumber',
+    'AU-ABN': 'registrationNumber',
+    'PY-PGN': 'classification',
```
pudo, line 14: is there a source for these somewhere?

followthemoney/cli/ocds.py
uhhhuh: Yes.

followthemoney/cli/ocds.py
uhhhuh: I have to consult the source data here (concerned about possible dupes).

followthemoney/cli/ocds.py
uhhhuh: http://org-id.guide/results?structure=all&coverage=all&sector=all

pudo: Hm, I don't think we want to drag in all that complexity. Maybe let's just add a default to `IDENTIFIERS.get` for `registrationNumber`?

PR #211, opened 2020-04-15
Updated defaultModel

kjacks:


pudo:
Weirdly enough, this seems to have Py and JS out of sync. I'm taking the liberty to do it on master....

PR #218, opened 2020-04-24
Add procedure to Contract schema and description to Interval

sunu:


followthemoney/schema/Interval.yaml
```diff
@@ -15,6 +15,9 @@ Interval:
     summary:
       label: "Summary"
       type: text
+    notes:
```
pudo, line 18: Do we really need another "misc" field? We've got like 4 of them on `Thing` and it's a total mess :/

PR #235, opened 2020-04-28
Add typing

Rafiot:
As of right now, it is totally useless, I just want to have a draft PR going to keep track of the merge conflicts.


followthemoney/types/registry.py
```diff
@@ -8,43 +10,40 @@ class Registry(object):
...
         return {t.group: t for t in self.types if t.group is not None}
 
-    def get(self, name):
+    def get(self, name: Union[str, PropertyType]) -> PropertyType:
         """For a given property type name, get its handling object."""
         # Allow transparent re-checking.
         if isinstance(name, PropertyType):
             return name
-        return self._types.get(name)
+        return self._types[name]
```
pudo, line 42: This is behaviour change. Can this be a separate PR?

followthemoney/compare.py
```diff
@@ -26,22 +31,22 @@
 }
 
 
-def compare(model, left, right):
+def compare(model: Model, left: Mapping, right: Mapping) -> float:
```
pudo, line 34: ... Union[Mapping,EntityProxy]

followthemoney/export/csv.py
```diff
@@ -1,26 +1,38 @@
...
-    def _open_csv_file(self, name):
+    def _open_csv_file(self, name) -> Tuple[TextIO, Any]:
         self.directory.mkdir(parents=True, exist_ok=True)
         file_path = self.directory.joinpath('{0}.csv'.format(name))
         handle = open(file_path, mode='w')
         writer = csv.writer(handle, dialect=self.dialect)
         return handle, writer
 
-    def _get_writer(self, schema):
+    def _write_header(self, writer, schema: Schema):
```
pudo, line 27: Does this apply 1:1 for the Neo4J Bulk exporter, the other user of the CSVMixIn?

followthemoney/graph.py
```diff
@@ -11,26 +16,31 @@ class Node(object):
...
-        self.type = type_
-        self.value = value
-        self.id = type_.node_id_safe(value)
-        self.proxy = proxy
-        self.schema = schema if proxy is None else proxy.schema
+    def __init__(self, type_: PropertyType, value: str,
+                 proxy: Optional[EntityProxy]=None,
+                 schema: Optional[Schema]=None):
+        self.type: PropertyType = type_
+        self.value: Optional[str] = value
```
pudo, line 23: Is value optional? That would then be when we get into the whole optional node ID mess.

followthemoney/graph.py
```diff
@@ -11,26 +16,31 @@ class Node(object):
...
+        self.schema: Optional[Schema] = schema if proxy is None else proxy.schema
 
     @property
-    def is_entity(self):
+    def is_entity(self) -> bool:
         return self.type == registry.entity
 
     @property
-    def caption(self):
+    def caption(self) -> Optional[str]:
```
pudo, line 36: See: is value optional?

followthemoney/graph.py
```diff
@@ -39,7 +49,9 @@ def to_dict(self):
         }
 
     @classmethod
-    def from_proxy(cls, proxy):
+    def from_proxy(cls, proxy: EntityProxy) -> 'Node':
+        if proxy.id is None:
+            raise Exception('<FIXME>')
```
pudo, line 54: FIXME :)

followthemoney/graph.py
```diff
@@ -60,50 +72,59 @@ class Edge(object):
...
             self.schema = proxy.schema
 
     @property
-    def source(self):
+    def source(self) -> Optional[Node]:
         return self.graph.nodes.get(self.source_id)
 
     @property
-    def source_prop(self):
+    def source_prop(self) -> Optional[Property]:
```
pudo, line 97: Edge must have a source prop, if you want we can check if it's a valid edge before.

followthemoney/graph.py
```diff
@@ -60,50 +72,59 @@ class Edge(object):
...
+        self.schema: Optional[Schema] = None
         if prop is not None:
             self.weight = prop.specificity(value)
         if proxy is not None:
             self.id = f"{source.id}<{proxy.id}>{target.id}"
             self.schema = proxy.schema
 
     @property
-    def source(self):
+    def source(self) -> Optional[Node]:
```
pudo, line 93: Not really optional.

followthemoney/graph.py
```diff
@@ -60,50 +72,59 @@ class Edge(object):
...
         return self.prop
 
     @property
-    def target(self):
+    def target(self) -> Optional[Node]:
         return self.graph.nodes.get(self.target_id)
 
     @property
-    def target_prop(self):
+    def target_prop(self) -> Optional[Property]:
```
pudo, line 108: This, however, is optional.

followthemoney/graph.py
```diff
@@ -129,39 +150,41 @@ class Graph(object):
...
         if node.id not in self.nodes:
             self.nodes[node.id] = node
         return self.nodes[node.id]
 
-    def _add_edge(self, proxy, source, target):
-        source = self._get_node_stub(proxy.schema.source_prop, source)
-        target = self._get_node_stub(proxy.schema.target_prop, target)
-        edge = Edge(self, source, target, proxy=proxy)
+    def _add_edge(self, proxy: EntityProxy, source: str, target: str):
+        if proxy.schema.source_prop is None or proxy.schema.target_prop is None:
```
pudo, line 180: I don't think either of these checks are needed given how it's invoked.

followthemoney/helpers.py
```diff
@@ -61,16 +66,17 @@ def entity_filename(proxy, base_name=None, extension=None):
...
     names aren't special and may not always be the caption."""
     if entity.schema.is_a('Thing'):
         names = entity.get('name')
         if len(names) > 1:
-            name = registry.name.pick(names)
-            if name in names:
-                names.remove(name)
-            entity.set('name', name)
-            entity.add('alias', names)
+            if isinstance(registry.name, NameType):
```
pudo, line 76: nice catch!

followthemoney/model.py
```diff
@@ -11,11 +13,11 @@
...
-    def __init__(self, path):
-        self.path = path
-        self.schemata = {}
-        self.properties = set()
-        self.qnames = {}
+    def __init__(self, path: str):
+        self.path: str = path
+        self.schemata: Dict[str, Schema] = {}
+        self.properties: Dict[str, Property] = {}
+        self.qnames: Dict[str, str] = {}
```
pudo, line 20: is qnames str not property?

followthemoney/model.py
```diff
@@ -62,49 +64,59 @@ def get_type_schemata(self, type_):
...
         """
-        left = self.get(left) or self.get(right)
-        right = self.get(right) or self.get(left)
-        if left.is_a(right):
-            return left
-        if right.is_a(left):
-            return right
-        common = list(left.schemata.intersection(right.schemata))
+        left_schema = self.get(left) or self.get(right)
+        if left_schema is None:
```
pudo, line 87: Behaviour change, the previous was to return the opposing side's schema, not throw. Can we make this a separate PR?

followthemoney/model.py
```diff
@@ -62,49 +64,59 @@ def get_type_schemata(self, type_):
...
             if schema.is_a(specific):
                 specific = schema
         return specific
 
-    def make_entity(self, schema, key_prefix=None):
+    def make_entity(self, schema: Mapping, key_prefix=None) -> EntityProxy:
         return EntityProxy(self, {'schema': schema}, key_prefix=key_prefix)
 
-    def get_proxy(self, data, cleaned=True):
+    def get_proxy(self, data: Mapping, cleaned=True) -> EntityProxy:
```
pudo, line 116: Union[Mapping, EntityProxy]

followthemoney/model.py
```diff
@@ -62,49 +64,59 @@ def get_type_schemata(self, type_):
...
+            raise InvalidData(msg % (left_schema, right_schema))
 
         specific = common[0]
         for schema in common[1:]:
             if schema.is_a(specific):
                 specific = schema
         return specific
 
-    def make_entity(self, schema, key_prefix=None):
+    def make_entity(self, schema: Mapping, key_prefix=None) -> EntityProxy:
```
pudo, line 113: Union[str, Schema], no?

followthemoney/namespace.py
```diff
@@ -69,14 +77,19 @@ def apply(self, proxy):
             if prop.type != registry.entity:
                 continue
             for value in signed.pop(prop):
-                value = get_entity_id(value)
-                signed.add(prop, self.sign(value))
+                _value = get_entity_id(value)
+                signature = self.sign(_value)
+                if signature:
+                    signed.add(prop, signature)
+                else:
```
pudo, line 84: signed.add handles None transparently, no need to bother with it here.

followthemoney/property.py
```diff
@@ -1,36 +1,36 @@
...
-        self.stub = data.get('stub', False)
+        self._label: str = data.get('label', name)
+        self._description: Optional[str] = data.get('description')
+        self.hidden: bool = data.get('hidden', False)
+        self.stub: bool = data.get('stub', False)
 
-        type_ = data.get('type', 'string')
-        self.type = registry.get(type_)
-        if self.type is None:
-            raise InvalidModel("Invalid type: %s" % type_)
```
pudo, line 31: See registry, weird behaviour change.

followthemoney/proxy.py
```diff
@@ -276,7 +292,7 @@ def __eq__(self, other):
         return self.id == other.id
 
     @classmethod
-    def from_dict(cls, model, data, cleaned=True):
+    def from_dict(cls, model: Model, data, cleaned: bool=True):
```
pudo, line 295: data: Union[Mapping, EntityProxy]

followthemoney/types/registry.py
Rafiot: The reason I changed it is because using `get` turns the return type to `Optional[PropertyType]`... and everything that calls `get` or `registry[<name>]` can be `None`.

So I can change that, but that will also mean a lot of type checking down the line:

```
$ (git::typing) mypy . 
followthemoney/types/registry.py:46: error: List comprehension has incompatible type List[Optional[PropertyType]]; expected List[PropertyType]
followthemoney/schema.py:117: error: Item "None" of "Optional[PropertyType]" has no attribute "name"
followthemoney/proxy.py:186: error: Argument 1 to "get_type_values" of "EntityProxy" has incompatible type "Optional[PropertyType]"; expected "PropertyType"
followthemoney/proxy.py:202: error: Item "None" of "Optional[PropertyType]" has no attribute "rdf"
followthemoney/proxy.py:224: error: Argument 1 to "get_type_values" of "EntityProxy" has incompatible type "Optional[PropertyType]"; expected "PropertyType"
followthemoney/proxy.py:228: error: Argument 1 to "get_type_values" of "EntityProxy" has incompatible type "Optional[PropertyType]"; expected "PropertyType"
followthemoney/property.py:32: error: Incompatible types in assignment (expression has type "Optional[PropertyType]", variable has type "PropertyType")
followthemoney/namespace.py:24: error: Item "None" of "Optional[PropertyType]" has no attribute "clean"
followthemoney/graph.py:55: error: Argument 1 to "Node" has incompatible type "Optional[PropertyType]"; expected "PropertyType"
followthemoney/compare.py:20: error: Dict entry 0 has incompatible type "Optional[PropertyType]": "int"; expected "PropertyType": "float"
followthemoney/compare.py:21: error: Dict entry 1 has incompatible type "Optional[PropertyType]": "int"; expected "PropertyType": "float"
followthemoney/compare.py:22: error: Dict entry 2 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:23: error: Dict entry 3 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:24: error: Dict entry 4 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:25: error: Dict entry 5 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:26: error: Dict entry 6 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:27: error: Dict entry 7 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:28: error: Dict entry 8 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:29: error: Dict entry 9 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/compare.py:30: error: Dict entry 10 has incompatible type "Optional[PropertyType]": "float"; expected "PropertyType": "float"
followthemoney/export/graph.py:9: error: Item "None" of "Optional[PropertyType]" has no attribute "name"
Found 21 errors in 8 files (checked 101 source files)
```

followthemoney/compare.py
Rafiot: Would it make sense to make `EntityProxy` inherit on `Mapping` then? because right now, it is not possible to call `dict(EntityProxy)`:
https://github.com/alephdata/followthemoney/blob/91e56b450a9c1bc29221e40cfc2512dfd1f97dfd/followthemoney/proxy.py#L25

followthemoney/export/csv.py
Rafiot: .... I realized that I missed adding typing in the neo4j exporter, so I'm pretty sure it doesn't vaguely work. I'll look at that.

followthemoney/graph.py
Rafiot: It doesn't have to, but then `EntityType.caption` has to return a string, and not `None`: https://github.com/alephdata/followthemoney/blob/91e56b450a9c1bc29221e40cfc2512dfd1f97dfd/followthemoney/types/entity.py#L32

followthemoney/graph.py
Rafiot: Yeah, I wasn't sure which kind of exception I should be using ;)

followthemoney/graph.py
Rafiot: So should I do `self.graph.nodes.get[self.source_id]` instead?

followthemoney/graph.py
Rafiot: I should do the following then:
1. `self.schema.source_prop.reverse` cannot be `None` (given `self.schema.source_prop` exists)
2. `self.prop` cannot be `None`

Rafiot: Okay, I missed a thing. In order to initialize an edge, we need to have at least a `prop` OR a `proxy` (to initialize the schema)

followthemoney/helpers.py
Rafiot: Thanks ;) mypy is pretty good at it

followthemoney/graph.py
Rafiot: I can add a type ignore on both calls to `self._get_node_stub`, but as `proxy.schema.source_prop` and `proxy.schema.target_prop` can be `None`, mypy will not like it if I don't do the test first.

followthemoney/model.py
Rafiot: woops, good catch... I changed the type of `self.properties` and `generate` is broken right now.

followthemoney/model.py
Rafiot: Sure, but the current code fails at runtime if either `left` or `right` are `None`: 
https://github.com/alephdata/followthemoney/blob/91e56b450a9c1bc29221e40cfc2512dfd1f97dfd/followthemoney/model.py#L86
https://github.com/alephdata/followthemoney/blob/91e56b450a9c1bc29221e40cfc2512dfd1f97dfd/followthemoney/model.py#L88

followthemoney/model.py
Rafiot: For that, `EntityProxy` need to inherit `Mapping`

followthemoney/model.py
Rafiot: woops, yes

followthemoney/namespace.py
Rafiot: :+1:

followthemoney/property.py
Rafiot: Yeah, I did that because get now always returns a `PropertyType`, or fails. See comment over there.

followthemoney/proxy.py
Rafiot: Same, we can do that if `EntityProxy` inherits  `Mapping`

followthemoney/compare.py
Rafiot: ... or at least I'm super confused how that is supposed to work right now.

Based on the doc, I cannot see how you can pass an `EntityProxy` to `dict` -> https://docs.python.org/3.8/library/stdtypes.html#dict

followthemoney/export/csv.py
Rafiot: So, I checked, yes, it will works. the method `_write_header` moved from `CSVExporter` to `CSVMixin` is overwritten in `Neo4JCSVExporter`

Rafiot:
Alright, I started to play with the code... and we (obviously) have circular imports all over the place. I'll figure something out.

Rafiot:
(rebased to HEAD)

Rafiot:
The test cases are passing again :)

pudo:
Wow, that's a lot of work you've done on this already. A bit more complex than I thought, I'll admit :) 

I've been doing a bit of work on making the whole thing pass a linter on this branch here: [origin/typing](https://github.com/alephdata/followthemoney/tree/typing).

The circular imports issue is really weird and leaves me very uncomfortable, would it be helpful is we broke compatibility below 3.7 (I hear theres a `__future__` that might be helpful)?

Another thing that I cannot quite wrap my head around yet is the method signatures on the type system - if I break up a method signature into two lines, `mypy` doesn't seem to accept the `type: ignore` any more, and breaks on differences between the upstream signature and the local one. Does this mean we can either have a happy linter or a happy type check or is there some workaround to this?

Rafiot:
Breaking compatibility below 3.7 is totally cool with me. I cannot force that before the end of this year on (Py)MISP side, because my users would kill me, but FTM will never be a required dependency, so that's definitely not a problem.

I think (?) all the circular imports are solved by using the `TYPE_CHECKING` trick. If I understand the documentation right, relying on `from __future__ import annotations` will only make the typing slightly nicer by avoiding using the quotes for classes defined in the same file: https://mypy.readthedocs.io/en/stable/common_issues.html#import-cycles
I'm not a fan of the `TYPE_CHECKING` trick, I find it pretty dirty, but for what I understand that's the only way to do that so mypy doesn't fail when we have different files needing definitions from each other.

And for the linter vs typing, this is the only trick I found: https://mypy.readthedocs.io/en/stable/common_issues.html#silencing-linters 
But I find it pretty dirty, unpleasant to read, and it doesn't seem to work on multiple lines: https://github.com/python/mypy/issues/1032 (this is absolutely disgusting and my linter absolutely hates that, but mypy is happy... But it seems to be supported properly with Python 3.8)

So for what I see right now, the remaining things that are failing with mypy are: 
1. incompatible with supertype (x2) -> linter vs. mypy 
2. `"EntityProxy" has incompatible type "Union[EntityProxy, Mapping[Any, Any]]"; expected "Mapping[Any, Any]"` -> I still don't understand how it is possible to have a `data` of type `EntityProxy` passed to a `dict()` (https://github.com/alephdata/followthemoney/blob/typing/followthemoney/proxy.py#L37) In my books, it only works if `EntityProxy` is an `Iterator`
3. `_get_node_stub" of "Graph" has incompatible type "Optional[Property]"; expected "Property"` because `Schema.source_prop` and `Schema.target_prop` can return `None`.

So yeah, looks good to me!

Rafiot:
Re. circular imports: https://github.com/alephdata/followthemoney/pull/246

Rafiot:
Part of https://github.com/alephdata/followthemoney/tree/typing

PR #246, opened 2020-05-10
Use __future__, fix tests

Rafiot:
Only if you're cool with making ftm python 3.7+ only, but that would fix the circular imports.

### DIFF
--- a/followthemoney/util.py
+++ b/followthemoney/util.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import os
 import logging
 from threading import local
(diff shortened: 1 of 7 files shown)


followthemoney/util.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import os
 import logging
 from threading import local
-from rdflib import Identifier  # type: ignore
+from rdflib.term import Identifier  # type: ignore
```
Rafiot, line 6: Not sure what happened there, but it was failing for me. Did the library change or something?

PR #249, opened 2020-05-25
Namespace

kjacks:
- Creates a namespace class to sign entity ids
- Adds namespace as an optional parameter to model.createEntity


pudo:
This looks to me like it's doing the right thing. Should we merge?

PR #263, opened 2020-06-02
new: MISP exporter

Rafiot:
FTM -> MISP, with test case.


pudo:
This looks great. I've noticed PyMISP from PyPI is 400kB, so let's add that as a normal dependency instead of doing some sort of additional dance?

Rafiot:
Yep, here it is :)

Issue #280, opened 2020-07-13
URLs miss the fragment after '#'


uhhhuh:
urlnormalizer.normalize_url returns urls with the part that follows "#" dropped.
"https://a.b.c/index.jsp#/ROC/companyDetailsRO.do?action=companyDetails&companyId=123" becomes "https://a.b.c/index.jsp"

simonwoerpel:
I am not sure if it is a bug or there are reasons why it is implemented with dropping the fragments, but i opened a PR that could solve this issue: https://github.com/alephdata/followthemoney/pull/282

pudo:
Adopted, closing.

PR #282, opened 2020-07-17
Don't drop url fragments?

simonwoerpel:
As it pops up from time to time (e.g. https://github.com/alephdata/followthemoney/issues/280) i am opening this here for discussion.

I don't know if this is a "bug" or for any other reason an intended feature?


pudo:
@dependabot squash and merge

PR #283, opened 2020-07-20
Additional properties to support alternate, translated text for documents

dschulz-pnnl:


pudo:
Thanks for the PR. Two comments: 

a) Can you explain a bit more the need for `altTextSrcLanguage` as opposed to `detectedLanguage` and `language`? 

b) If I could propose a naming, I'd be more explicit and go for `translatedText` and `translatedLanguage` - `alt*` is a bit ambiguous.

dschulz-pnnl:
a) The Google translation API provides the language it detected in its response. Based on the quality of their translation service, and the breadth of languages/dialects they support, the language they detect may differ from, and be more accurate than, the language detected during ingest.

b) The “alt” in the property names was suggested by @pudo in a slack thread where my colleague inquired about translation.

dschulz-pnnl:
@pudo, wondering if there is anything I can do to move this PR and its sibling (https://github.com/alephdata/aleph/pull/1278) forward. If the property names are holding it up, then I can see about getting them changed.

pudo:
I'm very sorry for delaying this PR, been distracted on other fronts. I would vote for the more explicit property names if possible, simply to avoid confusion over the purpose of the fields.

dschulz-pnnl:
Is there a way to remap a property name for data that has already been ingested? We have data that is already translated that I would like to move over to the "translatedText" property name.

pudo:
No, I don't think we've done this before. Might be worth looking into a postgres query against the ftm-store data, followed by a re-index?

dschulz-pnnl:
The property names have been changed from "altBodyText" to "translatedText".

pudo:
thanks a bunch!

PR #342, opened 2020-11-17
Multi Part Proxy

mynameisfiber:
This creates a new EntityProxy subtype that transparently represents multiple underlying proxies. It works by containing a list of constituent proxies, in addition to a list of proxies that definitely don't belong to it. This will serve as a representation for profiles in aleph.

### DIFF
--- a/followthemoney/multi_part_proxy.py
+++ b/followthemoney/multi_part_proxy.py
@@ -84,9 +84,13 @@ class MultiPartProxy(EntityProxy):
 
     def to_dict(self):
         return {
+            "id": self.id,
             "schema": self.schema.name,
             "parts": self.parts,
             "anti_parts": self.anti_parts,
             "proxies": self.proxies,
             "anti_proxies": self.anti_proxies,
         }
+
+    def __repr__(self):
+        return "<MPP(%r,%r)>" % (self.id, str(self))
--- a/followthemoney/multi_part_proxy.py
+++ b/followthemoney/multi_part_proxy.py
@@ -1,0 +1,92 @@
+from followthemoney.exc import InvalidData
+from followthemoney.proxy import EntityProxy
+
+
+class MultiPartProxy(EntityProxy):
+    def __init__(self, id=None, proxies=None, anti_proxies=None):
+        self.id = id
+        self.proxies = set(proxies or [])
+        self.anti_proxies = set(anti_proxies or [])
+        self._golden_proxy = None
+        self._build()
+
+    def _build(self):
+        if not self.proxies:
+            return
+        _golden_proxy = None
+        for proxy in self.proxies:
+            if _golden_proxy is None:
+                _golden_proxy = proxy.clone()
+            _golden_proxy.merge(proxy)
+        self._golden_proxy = _golden_proxy
+
+    def _merge_new_proxy(self, proxy):
+        if self._golden_proxy is None:
+            self._build()
+        self._golden_proxy.merge(proxy)
+
+    @property
+    def golden_proxy(self):
+        if self._golden_proxy is not None:
+            return self._golden_proxy.clone()
+        return None
+
+    @property
+    def schema(self):
+        return self._golden_proxy.schema
+
+    @property
+    def _properties(self):
+        return self._golden_proxy._properties
+
+    @property
+    def parts(self):
+        return [p.id for p in self.proxies]
+
+    @property
+    def anti_parts(self):
+        return [p.id for p in self.anti_proxies]
+
+    def add(self, prop, values):
+        raise NotImplementedError
+
+    def merge(self, other):
+        self.id = self.id or other.id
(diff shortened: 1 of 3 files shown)


followthemoney/proxy.py
```diff
@@ -264,11 +264,11 @@ def __len__(self):
         return self._size
 
     def __hash__(self):
-        return hash(self.id)
+        return hash(self.id or id(self))
 
     def __eq__(self, other):
         try:
-            return self.id == other.id
+            return self.id == other.id and not (self.id is None and other.id is None)
```
pudo, line 271: The overrides for `__eq__` and `__hash__` make me a bit nervous. The one for `__eq__` is in particular risky. Can we maybe override these in the subclass instead and then make explicit reference to `.parts`?

followthemoney/multi_part_proxy.py
```diff
@@ -0,0 +1,92 @@
...
+
+class MultiPartProxy(EntityProxy):
+    def __init__(self, id=None, proxies=None, anti_proxies=None):
+        self.id = id
+        self.proxies = set(proxies or [])
+        self.anti_proxies = set(anti_proxies or [])
+        self._golden_proxy = None
+        self._build()
+
+    def _build(self):
```
pudo, line 13: Just an alternative idea: what if we added a `clear()` method to the EntityProxy that flushes it's props and then just used the actual `self` for the storage, while keeping `add()`, `set()`, `pop()` and `remove()` overwritten? So `_build` would `clear`, then re-fill `self`....

followthemoney/multi_part_proxy.py
```diff
@@ -0,0 +1,92 @@
...
+        self.anti_proxies.discard(proxy)
+        if proxy in self.proxies:
+            self.proxies.discard(proxy)
+            self._build()
+
+    def __contains__(self, proxy):
+        return proxy in self.proxies
+
+    def to_dict(self):
+        return {
```
pudo, line 86: probably want `id`, too.

pudo (commented):
See the comment on whether we should use `self` instead of `golden_proxy`... 

I'm also wondering if you could assume some stewardship of the contents of `followthemoney.dedupe` and either get rid of them or adapt them. The original idea there was to be able to use the JSON responses of the `linkages` API and to feed them into the CLI version of FtM to perform entity re-writes (merging itself would be handled by an aggregator like ftmstore). But it didn't have semantics to name the constituent parts, which this does. On the plus side, it was able to then be fed into the `followthemoney.graph` code, which does NX/Neo4J etc.

followthemoney/multi_part_proxy.py
mynameisfiber: That's what I had wanted to do but there are multiple issues that make the implementation pretty dirt, most annoyingly:

1. the clear becomes a very specific to the multi_part_proxy in that it should probably keep context and existing IDs
2. many methods, including merge, break because `add` is not implemented. 

This was one of the motivators behind me questioning whether this object should be inheriting from `EntityProxy` at all as opposed to being a container.

followthemoney/proxy.py
mynameisfiber: These changes were made so that we can have a set of `EntityProxy`'s. As it stands, the current implementation seems flawed because two objects with completely different data but no ID set will be considered the same. 

ie; without the change:

```python
In [1]: from followthemoney import model
RDFLib Version: 5.0.0

In [2]: a = model.make_entity('Person'); a.add("name", "micha")

In [3]: b = model.make_entity('Person'); b.add("name", "pudo")

In [4]: set([a, b])
Out[4]: {<E(None,'micha')>}
```

with the change:

```python
In [1]: from followthemoney import model
RDFLib Version: 5.0.0

In [2]: a = model.make_entity('Person'); a.add("name", "micha")

In [3]: b = model.make_entity('Person'); b.add("name", "pudo")

In [4]: set([a, b])
Out[4]: {<E(None,'micha')>, <E(None,'pudo')>}
```

I'm fine changing this back if you are concerned, but the MultiPartProxy implementation will need some re-tuning or some documentation around this counter intuitive behavior.

mynameisfiber:
> The original idea there was to be able to use the JSON responses of the `linkages` API and to feed them into the CLI version of FtM to perform entity re-writes (merging itself would be handled by an aggregator like ftmstore).

Could you show me a stub of the CLI commands you want this new version to work with? I'd assume it'd go something like this:

1) Do several `GET /api/2/entitysets/{entityset_id}/items`requests and save to {entityset_id}.json
2) `ftm link -i {entityset_id}.json -i {other_entityset_id}.json > golden_entities.json`
3) `ftm export-* -i golden_entities.json`

Also, some general questions:

1. Are there any processes depending on `ftm.cli.dedupe.*` currently?
2. What was the intended functionality of `match-entities`/`match-decide`/`link`? Where were they getting their inputs from and what were they designed to feed?

PR #378, opened 2021-03-18
Major followthemoney.compare improvements

mynameisfiber:


mynameisfiber:
These changes seem to perform very well on aleph data. Comparing the distribution of scores across hand labeled data, we can see that `compare2` gives much better separation than `compare`.

image

versus

image

Another added benefit is that the scores should actually be on the interval [0, 1] now.

followthemoney/compare2.py
```diff
@@ -0,0 +1,103 @@
...
+            seen.add(fp)
+            yield fp
+
+
+def compare_names(left, right):
+    result = 0
+    left_list = list(_normalize_names(left.names))
+    right_list = list(_normalize_names(right.names))
+    if not (left_list or right_list):
+        raise ValueError
```
pudo, line 87: That doesn't seem like an error to me, just an edge case. Return `0` instead?

followthemoney/compare2.py
```diff
@@ -0,0 +1,103 @@
...
+        similarity = fuzz.WRatio(left, right, full_process=False) / 100.0
+        result = max(result, similarity)
+    return result
+
+
+def compare_countries(left, right):
+    left = left.country_hints
+    right = right.country_hints
+    if not (left or right):
+        raise ValueError
```
pudo, line 100: Same as above.

setup.py
```diff
@@ -30,6 +30,7 @@
         "click >= 7.0",
         "stringcase >= 1.2.0, < 2.0.0",
         "requests >= 2.21.0",
+        "fuzzywuzzy[speedup] >= 0.18.0",
```
pudo, line 33: Does the `[speedup]` contain lots of C that will make ftm less likely to run on people's computers? We're already slightly fucked on that front because of `pyicu`, just to check...

pudo (approved):
This looks good to me, jaro was there more on a hunch. Should we then make it replace `compare.py` instead?

setup.py
mynameisfiber: No. It ensures that `python-Levenshtein` is installed.

followthemoney/compare2.py
mynameisfiber: We need to distinguish between a 0 compare value (currently a 0), only one element in the pair having the property currently a `None`) or neither of the elements having the property (currently a `ValueError`). Given that, is this an alright setup or would you like it done some other way?

PR #379, opened 2021-03-19
Add programme property to Project

simonwoerpel:


simonwoerpel:
just noticed that `Thing` already has a "program" property: https://github.com/alephdata/followthemoney/blob/master/followthemoney/schema/Thing.yaml#L72

so, closing this.

PR #409, opened 2021-05-14
Rosencrantz/documentation review

Rosencrantz:
@pudo This is a first parse at the ftm documentation. I've tweaked a few things but nothing that I believe alters the meaning of the documentation.

If I had one observation that might need further consideration it is that we should probably remove some of the colloquialisms. There's some language here that I think some people might find it hard to understand. I've updated this where I felt confident, but it will likely require another pass.

More generally, and more long term I think that there might be a need for us to 'find a voice' for our docs. This isn't an important thing right now, but something for us to think about in the future.

### DIFF
--- a/followthemoney/namespace.py
+++ b/followthemoney/namespace.py
@@ -3,12 +3,15 @@
 
 Entity ID namespaces are a security mechanism related to the Aleph search index.
 
-This is because Aleph allows the user (via mappings or the API) to mint arbitary
-entity IDs. Entity IDs that are controlled by the user and not the system are,
-of course, unusual. At the same time, this makes it possible to generate bulk data
-outside Aleph, and then load entities into the system as a continuous :ref:`streams`.
+Aleph allows the user (via mappings or the API) to create arbitary entity IDs.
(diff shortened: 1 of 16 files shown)


followthemoney/namespace.py
```diff
@@ -3,21 +3,24 @@
...
-of course, unusual. At the same time, this makes it possible to generate bulk data
-outside Aleph, and then load entities into the system as a continuous :ref:`streams`.
+Aleph allows the user (via mappings or the API) to create arbitary entity IDs.
+Entity IDs that are controlled by the user and not the system are unusual.
+However, this makes it possible to generate bulk data outside Aleph,
+and then load entities into the system as a continuous :ref:`streams`.
 
-Namespacing works around this by making each entity ID consists of two parts:
+The problem is that having user controlled entity IDs increases the chance
+of conflict in the search index.
```
Rosencrantz, line 12: @pudo It felt like something was missing here, so I've added this short paragraph that I hope explains it. Let me know if this isn't factually correct.

PR #424, opened 2021-06-09
Improve golden record name selection

pudo:
Try to select more culturally appropriate names for entities which have multiple names.

### DIFF
--- a/followthemoney/types/name.py
+++ b/followthemoney/types/name.py
@@ -49,7 +49,9 @@ class NameType(PropertyType):
             lookup[norm].append(value)
 
         norm = setmedian(normalised)
-        forms = lookup.get(norm)
+        forms = lookup.get(norm, [])
+        if len(forms) <= 1:
+            return first(forms)
         return setmedian(forms)
 
     def _specificity(self, value):
--- a/followthemoney/types/name.py
+++ b/followthemoney/types/name.py
@@ -1,7 +1,7 @@
 from banal import ensure_list, first
 from normality import slugify
 from normality.cleaning import collapse_spaces, strip_quotes
-from fuzzywuzzy import fuzz
+from fuzzywuzzy import fuzz  # type: ignore
 from Levenshtein import setmedian  # type: ignore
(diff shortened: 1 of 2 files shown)


tests/types/test_names.py
```diff
@@ -19,6 +19,8 @@ def test_pick(self):
         self.assertEqual(names.pick(values), "Banana")
         self.assertEqual(names.pick("Banana"), "Banana")
         self.assertEqual(names.pick([]), None)
+        values = ["Robert Smith", "Rob Smith", "Robert SMITH"]
```
Rosencrantz, line 22: Boys. Don't. Cry

followthemoney/types/name.py
```diff
@@ -1,7 +1,7 @@
 from banal import ensure_list, first
 from normality import slugify
 from normality.cleaning import collapse_spaces, strip_quotes
-from fuzzywuzzy import fuzz
+from fuzzywuzzy import fuzz  # type: ignore
```
Rosencrantz, line 4: @pudo Can we just remove this, or is it being used elsewhere outside of these changes?

Rosencrantz (approved):

followthemoney/types/name.py
pudo: We might want to ask @mynameisfiber about that. FtM used to use just `python-Levenshtein` but now we've adopted `fuzzywuzzy` which seems to wrap `python-Levenshtein`. I'm not quite sure what the value add is.

mynameisfiber: It's being used in `NameType.compare()`. It uses `fuzz.WRatio` which has heuristics to apply differently weighted levenshtein-like distances depending on the size difference of the strings. It makes a huge difference when dealing with "Joe Shmo" vs "Joe Jonathan Shmo the Third" or other partial prefix/suffix matches.

Rosencrantz: Thanks for confirming @mynameisfiber

Issue #426, opened 2021-06-22
Make Passport more general


pudo:
We have a `Passport` schema, which is relatively specific but in its description states "this can be some other type of identification". I'd like to make a separate `Identification` schema from which `Passport` inherits and which can be used for many other types of ID (e.g. ID cards, voter registration, etc.)

Rosencrantz:
This one is now done right @pudo?

pudo:
Aye! :)

PR #428, opened 2021-06-22
Identification schema

pudo:
This implements #426 with a perspective to eventually discard the `Passport` schema in a couple of versions.


Rosencrantz (approved):

Rosencrantz:
@pudo Can you take a look at these conflicts and resolve?

Rosencrantz (approved):

Rosencrantz (approved):

sunu:
@pudo If I understand correctly, this doesn't affect existing data and only the reverse property name changes at run time from `passport` to `identification`. Is that right?

pudo:
Yes, this change (unlike the one I'd like to propose for license) should be fully backward compatible. Of course it will generate another index in Aleph :/

sunu:
Got it, thanks!

PR #441, opened 2021-07-06
Adopt prefixdate library for parsing dates

pudo:
This would adopt the `prefixdate` library, which makes some of the logic of the existing date parsing more explicit. When applied, this should:

a) Speed up date processing, both for format-string based parsing and for normal validation. 
b) Truncate prefixes for minutes and seconds from times when they are `00`. 
c) Fix an apparent issue with time zone conversion in `to_datetime`.

### DIFF
--- a/followthemoney/types/date.py
+++ b/followthemoney/types/date.py
@@ -1,14 +1,12 @@
-import re
 import os
-import pytz
 import logging
+from prefixdate import parse, parse_format, Precision
 from rdflib import Literal  # type: ignore
 from rdflib.namespace import XSD  # type: ignore
-from datetime import datetime, date
 
 from followthemoney.types.common import PropertyType
 from followthemoney.util import defer as _
-from followthemoney.util import dampen, sanitize_text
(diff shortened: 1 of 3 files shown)


Rosencrantz (commented):
Alright, just so I understand what you're proposing here. You're looking to remove the big, ugly Regex with a library that does the same thing, and this is why most of this PR is removing code?

pudo:
Haha, I guess the point is that the library uses one fat regex to do a lot of the stuff that's happening in individual steps here (like that weird `CUT_ZEROES` thing)...

One thing that's a bit "complicated" is that `prefixdate` disregards unknown suffixes on dates that it does not understand. That means `2001BANANA` is a valid date, and will normalise to `2001`. That's needed to pass the existing ftm tests, which contain some weird examples. Maybe we could at some point adapt and get more strict.

pudo:
OK, the time issue is fixed.

Rosencrantz (approved):

PR #442, opened 2021-07-19
Introduce a Security

pudo:
A security in the financial sense: some form of debt, equity, derivatives. 

I subjectively don't think this clashes with `Debt` as a link type, but don't quite know how to make the relation more obvious, either.

### DIFF
--- a/followthemoney/schema/Address.yaml
+++ b/followthemoney/schema/Address.yaml
@@ -1,6 +1,6 @@
-# OSM: https://wiki.openstreetmap.org/wiki/Proposed_features/House_numbers/Karlsruhe_Schema#Tags 
+# OSM: https://wiki.openstreetmap.org/wiki/Proposed_features/House_numbers/Karlsruhe_Schema#Tags
 # Seen in the wild:
-# 
+#
 # - full
 # - remarks / note
 # - c/o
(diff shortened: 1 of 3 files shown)


followthemoney/schema/Address.yaml
```diff
@@ -55,7 +55,10 @@ Address:
       description: "Zip code or postcode."
     region:
       label: "Region"
-      description: "State, province or area."
+      description: "Also province or area."
+    state:
+      label: "State"
+      description: "State oder federal unit."
```
sunu, line 61: typo here?

sunu: nvm. it's German for or

pudo: oops, fixed!

followthemoney/schema/Sanction.yaml
```diff
@@ -18,7 +18,7 @@ Sanction:
     entity:
       label: "Entity"
       reverse:
-        name: sanctionEntity
+        name: sanctions
```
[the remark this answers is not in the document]
pudo: Not within the data itself, since it's a sort of virtual reverse property that doesn't ever get stored in the data. The only issue might be UI deep links to the "sanctions" tab of a person, which would now go to the person base page. I'd be very surprised if any links like that were in use.

Rosencrantz (approved):

PR #458, opened 2021-08-05
I  made these changes to get the existing tests to pass

Rosencrantz:
@sunu Seems to be ok given the docs and source that I checked. Would love your thoughts though!


followthemoney/export/rdf.py
```diff
@@ -14,10 +14,11 @@ def __init__(self, fh, qualified=True):
 
     def write(self, proxy, **kwargs):
         graph = Graph()
+
         for triple in proxy.triples(qualified=self.qualified):
             graph.add(triple)
         try:
             nt = graph.serialize(format="nt").strip()
-            self.fh.write(nt.decode("utf-8") + "\n")
+            self.fh.write(nt + "\n")
```
sunu, line 22: `nt` defaults to `ascii` encoding. So this will break for unicode text. May be we should consider using `nt11` and add a test for unicode strings.

setup.py
```diff
@@ -42,8 +42,8 @@
         "python-stdnum >= 1.16, < 2.0.0",
         "pantomime >= 0.4.0, < 1.0.0",
         "pytz >= 2021.1",
-        "rdflib >=5.0.0, < 5.1.0",
-        "networkx >= 2.5, < 2.7",
+        "rdflib >= 5.0.0, < 6.1.0",
```
sunu, line 45: version 6.x introduces breaking changes. Current changes will fail on version 5.x for example. So we should narrow the range down to only version 6.x.

sunu (changes requested):

followthemoney/export/rdf.py
Rosencrantz: Curious where you found nt11, doesn't seem to exist here: https://rdflib.readthedocs.io/en/stable/plugin_serializers.html?highlight=graph.serialize#id1 what am i missing?

setup.py
Rosencrantz: Good point. This reflects the dependabot change but I agree it makes no sense across a major version boundary

followthemoney/export/rdf.py
sunu: Seems like the docs are out of date. https://github.com/RDFLib/rdflib/blob/038af4547e799845e08986867b24acafcbe72d48/rdflib/plugin.py#L170 and https://github.com/RDFLib/rdflib/blob/038af4547e799845e08986867b24acafcbe72d48/rdflib/plugins/serializers/nt.py#L35

setup.py
sunu: FYI, your merge conflict fix removed the 5.x to 6.x change you made.

sunu (approved):

Rosencrantz:
Update the version of rdflib and tweak implementation to be compatible.

PR #477, opened 2021-09-03
Work on typing the types subsystem

pudo:
This is a first stab at introducing more type info to the ftm libraries. I'm trying to keep this incremental because by the very nature of it, it's going to touch a ton of files and go out of date quick :/


followthemoney/types/common.py
```diff
@@ -51,21 +55,33 @@ class PropertyType(object):
...
+        text = sanitize_text(raw)
         if text is None:
             return None
-        return self.clean_text(text, **kwargs)
+        return self.clean_text(text, fuzzy=fuzzy, format=format, proxy=proxy)
 
-    def clean_text(self, text: str, **kwargs) -> Optional[str]:
+    def clean_text(
+        self,
+        text: str,
```
Rosencrantz, line 80: And here too, if feasible. Make the argument names consistent across the common definition and the overrides.

followthemoney/types/identifier.py
```diff
@@ -1,4 +1,5 @@
 import re
+from typing import Callable, Sequence
```
Rosencrantz, line 2: This doesn't seem to be used

followthemoney/types/language.py
```diff
@@ -85,7 +91,7 @@ class LanguageType(EnumType):
     LANGUAGES = get_env_list("FTM_LANGUAGES", LANGUAGES)
```
Rosencrantz, line 91: This is marked as constant coz it's all uppercase, but we're overwriting it on the line below. Is this legit?

pudo:
I think I've got all the names in sync now. Opted for `value` for `validate()` because `value` gets used for all the other methods that accept an existing FtM prop value.

followthemoney/types/language.py
pudo: `constant-ish`. It's sort of stable at run-time, is what I mean.

Rosencrantz (approved):

pudo:
Merged via #479.

PR #478, opened 2021-09-10
Types, round 2

pudo:
This PR aims to type-annotate the key FtM schema object, like `Model`, `Schema`, `Property` and `EntityProxy` :) 

fc90a70c6abd68dd5d0954bcf7cff7df


Makefile
```diff
@@ -1,12 +1,20 @@
...
 test:
 	nosetests --with-coverage --cover-package=followthemoney --cover-erase --cover-html --cover-html-dir=coverage-report
 
+types:
+	mypy --strict followthemoney/types \
+		followthemoney/schema.py \
+		followthemoney/property.py \
+		followthemoney/proxy.py \
+		followthemoney/graph.py \
+		followthemoney/model.py
```
Rosencrantz, line 16: Is the plan here that we add more and more files to this list as they're converted to be typed?

followthemoney/graph.py
```diff
@@ -52,22 +66,22 @@ def to_dict(self):
...
+    def __repr__(self) -> str:
         return "<Node(%r, %r, %r)>" % (self.id, self.type, self.caption)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(self.id)
 
-    def __eq__(self, other):
-        return self.id == other.id
+    def __eq__(self, other: Any) -> bool:
```
Rosencrantz, line 83: Any?

followthemoney/property.py
```diff
@@ -1,21 +1,71 @@
...
+
+
+class PropertyToDict(PropertyDict, total=False):
+    name: str
+    qname: str
+    reverse: Optional[str]
+    stub: Optional[bool]
+
+
+class Property:
```
Rosencrantz, line 41: Might be worth defining a standard here as I've been putting the root type class at the top. Not bothered where it lives, but I do like consistency.

followthemoney/graph.py
pudo: That's a suggestion by mypy, pretty weird.

Makefile
pudo: Yep. It establishes an explicit coverage as long as the whole package isn't compliant.

Rosencrantz (approved):
This thing is massive and I'm not, in all honestly going to be able to give it an appropriate review - testament to the fact that changes should be small and incremental I guess - with the time I have. 

I'm going to make the assumption that, with tests passing etc that this change will work as expected. I'll likely live to regret this decision

pudo:
I can make a commitment to be around to help sort out the fallout visible in the open. For the datavault, let's just pray.

pudo:
Merged via #479.

PR #560, opened 2022-01-28
Tag companies as offshores

pudo:
This is maybe a bit provocative but could be a fun form of analysis.

### DIFF
--- a/followthemoney/offshore.py
+++ b/followthemoney/offshore.py
@@ -1,0 +1,49 @@
+from followthemoney.types import registry
+from followthemoney.proxy import EntityProxy
+
+# Derived from: https://fsi.taxjustice.net/en/introduction/fsi-results
+OFFSHORE_COUNTRIES = set(
+    (
+        "ky",
+        "ch",
+        "sg",
+        "lu",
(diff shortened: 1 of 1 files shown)


followthemoney/offshore.py
```diff
@@ -0,0 +1,49 @@
...
+
+def offshore_from_jurisdiction(proxy: EntityProxy) -> EntityProxy:
+    """Tag organizations linked to a well-known offshore jurisdiction as
+    offshores automatically. Complete generalization, use only in experiments."""
+    if not proxy.schema.is_a("Organization"):
+        return proxy
+    countries = set(proxy.get("country", quiet=True))
+    countries.update(proxy.get("jurisdiction", quiet=True))
+    if len(countries.intersection(OFFSHORE_COUNTRIES)) > 0:
+        proxy.add("topics", "corp.offshore")
```
sunu, line 48: Are you using any convention for the name of the topic string? Would be good to have a list of these built-in topic strings somewhere if we're going to start using them more and more.

Rosencrantz (approved):

PR #592, opened 2022-02-18
Make Thing.name type name

### LINKED ISSUE #526: Thing's name properties are just strings
It seems to me that a Thing's `name` property should be of `name` type and not `string` type. This way thing's names would be matchable by the various comparison operations.

Thoughts?

pudo:
Fixes #526

I find it sort of difficult to reason about the consequences of this, but it's probably not fatal. For example, it might make the titles of spreadsheet tabs show up in the `name` facet in search :)


Rosencrantz (approved):
Works for me.

PR #606, opened 2022-03-04
Introduce region codes for invaded regions of Ukraine

pudo:
This seems like the minimally offensive way of doing this: use administrative subdivision codes of the Ukrainian government. cf. https://en.wikipedia.org/wiki/ISO_3166-2:UA


Rosencrantz (approved):
Note to anyone looking at this PR in the future. The codes here are the relevant ISO region codes https://en.wikipedia.org/wiki/ISO_3166-2:UA

Issue #616, opened 2022-03-14
Adding a polyline property to Trip or other entity types.


brrttwrks:
As we are considering modeling a Trip in FtM, the thought occurred to me that, like an Address has a lat/lon, a Trip could/should have a polyline geometry to record not just start and end destinations, but the actual trajectory. I can imagine a number of cases where modeling this might be advantageous. One potential is to document a trip's trajectory - a flight, a vessels path from port to port, the path of a person (timeline?). Another might be to support more interesting UI components if we have map views in Aleph. 

I am throwing this out there, but I am sure I am overlooking a number of things that make this more complicated than I am imagining it :).

pudo:
I'd be a bit careful with randomly committing to support geo features in FtM. Doing so would be a significant (year+) amount of effort and I would be willing to bet a good case of beer that it would make Aleph totally collapse with complexity.

PR #620, opened 2022-03-16
Draft schema for a Trip to model flights etc

sunu:
How does this look? @brrttwrks @pudo @jlstro @Rosencrantz 

 Also, what do people think about adding a `parent` property to `Event` and `Trip` so that we can nest events and trips to represent a sequence of events as part of a larger event?

refs #616

### DIFF
--- a/followthemoney/schema/Trip.yaml
+++ b/followthemoney/schema/Trip.yaml
@@ -23,10 +23,18 @@ Trip:
   properties:
     startLocation:
       label: "Start location"
-      type: address
+      type: entity
+      range: Address
+      reverse:
+        name: tripsDeparting
+        label: "Trips departing"
     endLocation:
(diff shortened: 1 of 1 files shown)


pudo:
Nice work! I wonder if using `Address` objects on this may be a better path, so that those can be browsed as objects and show their "trips incoming", "trips departing"?

sunu:
Yes, that makes sense. I'll update the schema

brrttwrks:
I definitely agree with using `Address` objects.  If we are referencing, say, airports, that might be good to show which trips departed or arrived from that address. Also, agree with the `parent` property - I can definitely see use cases for it and it is a nice way to relate events to super-events or vice versa.

Rosencrantz (approved):

PR #669, opened 2022-05-06
Fix captions

Rosencrantz:


js/src/entity.ts
```diff
@@ -116,6 +116,10 @@ export class Entity {
   getCaption(): string {
     for (const property of this.schema.caption) {
       for (const value of this.getProperty(property)) {
+        if (typeof value !== "string" && typeof value.getCaption === "function") {
```
Rosencrantz, line 119: This change allows for captions to come from other entities. With trip startLocation, endLocation etc are Address entities so if only these are included without this change an error occurs.

pudo:
If you're going to add entity props as captions for entities, I'd recommend adding a check to this method that skips entity-typed props in the backend: 

https://github.com/alephdata/followthemoney/blob/master/followthemoney/proxy.py#L368-L371

Rosencrantz:
Not sure how I feel about the discrepancy between front and backend. In the UI we drill down to a string based caption, on the backend we default to the entity label. I think I may tweak it so that the frontend behaves like the backend

sunu (commented):
Given the changes in https://github.com/alephdata/followthemoney/pull/684, do we need to make these changes any more?

Rosencrantz:
> Given the changes in #684, do we need to make these changes any more?

Yes, I think we do. If this situation has happened once, it will happen again at some point. This change ensurers that when someone does decide to use an Entity as a caption, that it is handled more gracefully than it currently is.

sunu:
> Yes, I think we do. If this situation has happened once, it will happen again at some point. This change ensurers that when someone does decide to use an Entity as a caption, that it is handled more gracefully than it currently is.

I think in that case it would be better to raise an error rather than try to handle it

pudo:
@sunu cf. #689 would do that.

Rosencrantz:
> If you're going to add entity props as captions for entities, I'd recommend adding a check to this method that skips entity-typed props in the backend:

I don't disagree, as pudo mentions #689 deals with this in that way

Rosencrantz:
Closing this PR in favour of #689

PR #671, opened 2022-05-09
Remove ftm-enrich from ftm core module

pudo:
The `followthemoney-enrich` package is not actively used by Aleph and I'm not aware of any other consumers. I'm proposing to remove it from FtM, in part to reduce ambiguity as we begin to implement similar/related things in `nomenklatura`.


Rosencrantz:
@sunu @brrttwrks Can either of you comment on this at all, can we confirm that we are not using this functionality and that there are no big issues with bringing this change into FtM

PR #672, opened 2022-05-09
Remove record linkage functions

pudo:
This would remove the record-linkage related functions from FtM. These are not used in Aleph (as best I know) and their CLI use was historically limited to Run &D efforts within OCCRP that are now dead. 

WARNING: This will actually break Aleph code by moving `followthemoney.dedupe.judgement.Judgement` to `followthemoney.dedupe.Judgement`. I would actually suggest moving it to aleph entirely.


sunu:
@pudo As far as I can see, Aleph has its own copy of Judgement here: https://github.com/alephdata/aleph/blob/5663e30d9138c2ff3137caf40a9afbdca896223d/aleph/model/entityset.py#L17

Do we need to use a common implementation of Judgement?

PR #681, opened 2022-05-10
Idea: transform values via python modules in mapping

simonwoerpel:
```yaml
name:
  column: "name"
  apply_func: "fingerprints.generate"
```

```yaml
website:
  column: "url"
  apply_func: "myproject.cleaners:url"
```

This is a mvp draft, I just want to ask, does it make sense for you to develop this further?
Of course a proper implementation would need some safety checks for the python module import and so on...

In general, I'd love a feature like this ;)

On the other hand, one could argue that tasks like this should be done in data cleaning before applying a mapping.

### DIFF
--- a/followthemoney/mapping/property.py
+++ b/followthemoney/mapping/property.py
@@ -7,7 +7,7 @@ from banal import keys_values, as_bool
 from followthemoney.helpers import inline_names
 from followthemoney.exc import InvalidMapping
 from followthemoney.proxy import EntityProxy
-from followthemoney.util import get_entity_id, sanitize_text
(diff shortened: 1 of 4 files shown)


pudo:
Very cool idea, has me a bit scared because as it stands right now, it's an RCE vuln in Aleph (upload a list of file names, call a function to cat them into your entities)... I guess the minimum would be to define a safe/unsafe mode for mappings in which these would be enabled or disabled. 

More broadly, I do think that a lot of the uses for this might be data cleaning... maybe a more restrictive language like jinja could be an option?

PR #695, opened 2022-05-27
Similar schema for soft entity merges

pudo:
The philosophers have hitherto only interpreted the world in various ways. 

The point, however, is to change it.

### DIFF
--- a/followthemoney/schema/Similar.yaml
+++ b/followthemoney/schema/Similar.yaml
@@ -7,14 +7,12 @@ Similar:
   matchable: false
   generated: true
   featured:
-    - title
-    - fileName
-    - mimeType
-    - parent
+    - candidate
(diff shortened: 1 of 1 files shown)


Rosencrantz:
This looks fine to me. @brrttwrks any comments?

brrttwrks:
What is the fileName property representing? Other than that, looks OK. This looks like it will be useful more in data exchange - trying to think how it might be represented inside of Aleph, given xref gives its own score. I think we have one other entity type that doesn't inherit from either Thing/Interval -> Analyzable, but that makes sense here.

pudo:
`fileName` was a bit of code cruft, gone. And I agree this is not essential to Aleph really, unless people want to use it in diagrams ....

PR #741, opened 2022-06-20
Update README.md

DrWhax:
fix webvowl link with a domain that's not expired and can show webvowl.


Rosencrantz (approved):
@DrWhax Thanks for identifying and fixing this issue. Really appreciated.

PR #742, opened 2022-06-21
Proposed ftm 3.0

pudo:


Rosencrantz:
@pudo @brrttwrks @sunu There's a lot here. I can get behind making FtM more lean and removing unused features. I am going to merge this in so that we can test the implications for Aleph.

PR #744, opened 2022-06-23
Make Address a Thing

pudo:
Based on feedback from @ozhyrenkov here's a change to make an Address a Thing. Note that Address (in its common usage) does not describe the presence of another thing at this address, merely the concept of a place. We may need to sooner or later introduce an Address-to-Thing interval to say: "This company was headquartered at this Address from X to Y"....

### DIFF
--- a/followthemoney/schema/Address.yaml
+++ b/followthemoney/schema/Address.yaml
@@ -19,7 +19,7 @@ Address:
   label: Address
   plural: Addresses
   extends:
-    - Interval
+    - Thing
   description: >
     A location associated with an entity.
   matchable: true


ozhyrenkov:
Is it that easy? Will it cause any problems for the entities created in Alephs around the globe?

pudo:
Well doing it is that easy. I made the PR to try and see how destructive this would be on my local pipeline.

brrttwrks:
This seems like it will/should not be too destructive. There are other properties that are Things in similar way - parent for LegalEntity. Though I agree that having a startDate and endDate makes sense here, which is intervally. Curious how this will play out in the UI, though.

Rosencrantz (approved):

Issue #761, opened 2022-07-08
Proposed schema: Role/Post/Position


pudo:
This one is an odd one, so I'm pitching it here before implementing it in a PR. Basically, a lot of data sources I see at OpenSanctions contain the notion of someone holding an office (e.g. President of country X, member of parliament Y). Now, in theory a lot of this could be expressed as `Membership`, `Employment` or `Directorship`, but there's some issues: 

a) You don't always know which one of these it is from the data (e.g. is Biden an employee of the US, or a Member of it?)
b) It ends up leading to the creation of dumb stub entities, like "Republic of X", or "Commission to do Z" that exist only to express someone holding a position. 

So I'd like to propose making a new model for `Post`, in which the "far side" (Organization) is optional and can also be a string. Otherwise it'd be a pretty normal `Interval` linking to a `Person`. It's duplicative with the schemata mentioned above, but just allows expressing a different category of data.

What do people think?

PR #773, opened 2022-07-22
Post schema and helper method

pudo:
Explained in #761.

### DIFF
--- a/followthemoney/helpers.py
+++ b/followthemoney/helpers.py
@@ -6,7 +6,7 @@
 # If anyone were to swap out the default model, this would
 # probably be the first place to break.
 from os.path import splitext
-from typing import Iterable, List, Optional
+from typing import Iterable, List, Optional, Set
 from normality import safe_filename
 from mimetypes import guess_extension
 from itertools import product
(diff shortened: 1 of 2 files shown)


Rosencrantz (approved):

pudo:
I'm still not sure I like the `organization` and `organizationEntity` thing. Maybe these two want to have different names. 

Also: I wonder if we should use this to slowly deprecate `Employment`, which is sort of a subset of a post....

tillprochaska:
> I'm still not sure I like the organization and organizationEntity thing. Maybe these two want to have different names.

One thing to keep in mind when adding both a string and entity property is that it will likely confuse end users. It isn’t really intuitive to have two separate properties for basically the same thing, especially if both properties are labeled the same.

(In the Aleph UI, we currently do not surface the type of a property. Even if we did, it would still confuse some users, as not everyone is familiar with the underlying concepts of FtM and the differences between entity and string properties etc. See https://github.com/alephdata/aleph/issues/2350 for context.)

pudo:
I guess one way to resolve this and also draw a clearer separation from `Employment`, `Directorship`, and so on would be to get rid of the `organizationEntity` entirely.

PR #774, opened 2022-07-22
Remove dependency on fuzzywuzzy

pudo:
As proposed in #764.


pudo:
Are you confident that the results from these two comparisons will be the same?

_Originally posted by @Rosencrantz in https://github.com/alephdata/followthemoney/issues/776#issuecomment-1195093652_

No - the numbers would be different but I'm reasonably sure that their ordering will remain the same. So a closer match will still score higher than a less precise match.

PR #790, opened 2022-08-10
Added filter_recent command to cli to filter out old entities

manonromain:
When the store becomes large, and is partially updated regularly (eg if you have a quota for accessing an API - the French Company Registry is one example), it is useful to parse only recently added entities. To use `filter_recent`, the `retrievedAt` property of entities needs to be properly filled out.

### DIFF
--- a/followthemoney/cli/filter_recent.py
+++ b/followthemoney/cli/filter_recent.py
@@ -1,0 +1,108 @@
+import click
+import datetime
+from pathlib import Path
+from typing import Iterable, Optional
+
+from followthemoney import model
+from followthemoney.proxy import E, EntityProxy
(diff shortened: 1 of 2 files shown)


Rosencrantz:
Thanks for creating this pull request @manonromain, we're taking a look and will come back with questions if we have them.

Rosencrantz:
@manonromain Just thinking about this, perhaps a different way of looking at this would be to implement a more generic filter where you can provide a value that you want to filter and the argument by which you want to filter?

pudo:
Closing this as it's a bit too specific to include, and we might want to look at implementing a more general filtering tool in the future.

PR #825, opened 2022-09-14
feat: add a first flat version of Call For Tenders entity schema

Julm:


followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
+CallForTenders:
```
pudo, line 1: Would `Tender` also work? I'm just trying to keep the names simple, but this may be less precise.

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - Thing
+  required:
+    - title
+  featured:
+    - title
+    - contractingAuthority
+  caption:
+    - title
+  properties:
+    title:
```
pudo, line 15: I think `name` is defined on `Thing`, perhaps that works already?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+      label: "Title"
+    uid:
+      label: "CfT Unique ID"
+    cftCaUid:
+      label: "CfT CA Unique ID"
```
pudo, line 20: That seems super super specific. Really needed? 
Btw all ID fields should be `type: identifier`

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    contractingOnBehalfOf:
+      label: "Published on behalf of"
```
pudo, line 22: Should this be an entity reference to a LegalEntity?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    contractingAuthority:
+      label: "Name of Contracting Authority"
```
pudo, line 24: Should this be an entity reference to a LegalEntity?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    evaluationMechanism:
+      label: "Evaluation Mechanism"
+    description:
+      label: "Description"
```
pudo, line 28: There's `summary` on Thing, do we need `description`?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    directive:
+      label: "Directive"
+    procedure:
+      label: "Procedure"
+    involves:
+      label: "CfT Involves"
+    cpvCodes:
+      label: "CPV Codes"
+    eAuctionsIncluded:
+      label: "Inclusion of e-Auctions"
```
pudo, line 40: What does this mean?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    description:
+      label: "Description"
+    procurementType:
+      label: "Procurement Type"
+    directive:
+      label: "Directive"
+    procedure:
+      label: "Procedure"
+    involves:
+      label: "CfT Involves"
```
pudo, line 36: What does this mean?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+      label: "NUTS Code"
+      description: "Nomenclature of Territorial Units for Statistics (NUTS)"
+    aboveOrBelowThreshold:
+      label: "Above or Below threshold"
+    paymentOptions:
+      label: "Payment Options"
+    submissionDeadline:
+      label: "Submission deadline"
+    clarificationDeadline:
+      label: "End of clarification period"
```
pudo, line 51: `type: date`?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    euFunding:
+      label: "EU funding"
+    multipleTenders:
+      label: "Multiple tenders will be accepted"
+    publicationDate:
+      label: "Date of Publication/Invitation"
+    tedLinksForNotices:
+      label: "TED links for published notices"
+    dateOfAwarding:
+      label: "Date of Awarding"
```
pudo, line 69: maybe `awardDate` to get consistency? `type: date`.

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    fallsUnderGPPScope:
+      label: "Does this Call fall under the scope of GPP?"
```
pudo, line 71: GPP? Maybe add a description?

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,77 @@
...
+    certificationCheck:
+      label: "Certification check"
+    tenderers:  
+      label: "Tenderers"
+    awardedContracts:
+      label: "Awarded Contracts"
```
pudo, line 77: We do have an existing `ContractAward` schema, maybe we could just link to that?

pudo (commented):
Random thoughts....

followthemoney/schema/CallForTenders.yaml
Julm: I do understand but I don't think so; this is the process of calling companies (tenderers) to make an offer / a tender. And then, from the call, one contract (or more, depending on number of lots I think) is awarded to one or more tenderer(s).

followthemoney/schema/CallForTenders.yaml
Julm: I could use name instead of title (which is the field name of choice from the source)

followthemoney/schema/CallForTenders.yaml
Julm: I already asked the same question and it seems so. I'll change the type though yes, thanks.

followthemoney/schema/CallForTenders.yaml
Julm: It should eventually yes, but I'd like to make a flat version first, and when we agree on the whole model and I get the process for emitting it from a memorious scraper, I'll make a structured one if you may accept.

pudo:
Closing in favour of #867.

PR #835, opened 2022-09-30
Re-do documentation

pudo:
Just a little experiment in re-doing the documentation as an auto-generated JS site


Rosencrantz:
Killing this. @tillprochaska is working on a new documentation system based on Astro which should be available early next year.

Issue #811, opened 2022-09-02
Define which properties should be used to display an entity in a timeline


tillprochaska:
I’m currently thinking about enhancements for the timelines feature in Aleph. One common user request is to actually visualize timeline items. For example like this:

![timelines example](https://user-images.githubusercontent.com/1512805/188194957-f7bb6823-5dc7-4a84-9e37-c84fd407596b.png)

Every entity would be represented by a bar. The position and length of the bar depends on the start and end date of the entity. Which properties should be considered the start/end date of an entity can be different per entity.

It’s fairly simple for interval entities -- as they all have `date`/`startDate` and `endDate` properties. It’s not that simple in other cases, and there might be even multiple properties that could be used as an entites start/end date.

| Schema | start date | end date |
| - | - | - |
| `LegalEntity` | `incorporationDate` | `dissolutionDate` |
| `File` | `date`, `authoredAt`, `publishedAt` | |
| `Contract` | `contractDate` | |
| `CourtCase` | `fileDate` | `closeDate` |
| … | … | … |

I’m not 100% sure about this yet (and whether this should be part of the FtM model), but wanted to bring this up for discussion early. Also we already do something similar with the `edge` metadata.

Related:
* https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/temporal
* https://schema.org/temporalCoverage

pudo:
Proposed verbiage in the schema files:

```yaml
Test:
   extends: ....
   temporalExtent:
      start: 
        - startDate
        - date
      end:
         - endDate
         - date
```

tillprochaska:
I have written this done in a slightly more detailed form. Would love to hear feedback/comments. Any edge cases that aren’t covered by this? Does the naming make sense?

***

Schema YAML files support an new, optional key `temporalExtent` to store metadata about the properties that should be considered the temporal start and end of an entity:

```yml
Event:
  ...
  temporalExtent:
    start:
      - startDate
      - date
    end:
      - endDate
      - date
  ...
```

`start` and `end` are lists of properties, as in some cases multiple properties can be used as the start or end date when displaying an entity in a timeline. For example, a multi-day event might have the `startDate` and `endDate` properties set, while a single-day event might only have the `date` property filled out.

`start` and `end` properties must be of type `Date`.

If multiple `start` properties of an entity are set or if properties have multiple values, the smallest value should be considered the temporal start.

[...]

PR #849, opened 2022-10-14
Add temporal coverage metadata

tillprochaska:
This adds basic support for temporal extent metadata (#811) to the Python and TS libraries. It also adds helpers to entity proxies to retrieve the temporal start and end:

```python
company = model.make_entity('Company')
company.add('incorporationDate', '2021-01-01')
company.add('dissolutionDate', '2021-12-31')
company.temporal_start # '2021-01-01'
company.temporal_end # '2021-12-31'

event = model.make_entity('Event')
event.add('date', 2022-01-01')
company.temporal_start # '2021-01-01'
company.temporal_end # None
```

```ts
const company = model.createEntity('Company');
company.setProperty('incorporationDate', '2021-01-01');
company.setProperty('dissolutionDate', '2021-12-31');
company.getTemporalStart(); // '2021-01-01'
company.getTemporalEnd(); // '2021-12-31'

const event = model.createEntity('Event');
event.setProperty('date', '2021-01-01');
event.getTemporalStart(); // '2021-01-01'
event.getTemporalEnd(); // null
```

I also added temporal extent metadata to all existing schemata (where applicable), although I’d consider this work in progress and I’ll wait for feedback from people who have real-world experience working with the model.

I added the temporal extent metadata to the schemata documentation which is [available here](https://followthemoney--849.org.readthedocs.build/en/849/model.html#event).

### DIFF
--- a/followthemoney/proxy.py
+++ b/followthemoney/proxy.py
@@ -330,26 +330,28 @@ class EntityProxy(object):
         return self.get_type_values(registry.country)
 
     @property
-    def temporal_start(self) -> Optional[str]:
+    def temporal_start(self) -> Optional[Tuple[Property, str]]:
         """Get a date that can be used to represent the start of the entity in a timeline.
         If there are multiple possible dates, the earliest date is returned."""
         values = []
 
         for prop in self.schema.temporal_start_props:
-            values += self.get(prop.name)
+            values += [(prop, value) for value in self.get(prop.name)]
 
-        return next(iter(sorted(values)), None)
+        values.sort(key=lambda tuple: tuple[1])
+        return next(iter(values), None)
 
     @property
-    def temporal_end(self) -> Optional[str]:
+    def temporal_end(self) -> Optional[Tuple[Property, str]]:
         """Get a date that can be used to represent the end of the entity in a timeline.
         If therer are multiple possible dates, the latest date is returned."""
         values = []
 
         for prop in self.schema.temporal_end_props:
-            values += self.get(prop)
+            values += [(prop, value) for value in self.get(prop.name)]
 
-        return next(iter(sorted(values, reverse=True)), None)
+        values.sort(reverse=True, key=lambda tuple: tuple[1])
+        return next(iter(values), None)
 
     def get_type_inverted(self, matchable: bool = False) -> Dict[str, List[str]]:
         """Return all the values of the entity arranged into a mapping with the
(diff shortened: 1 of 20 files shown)


followthemoney/proxy.py
```diff
@@ -329,6 +329,28 @@ def countries(self) -> List[str]:
         """Get the set of all country-type values set of the entity."""
         return self.get_type_values(registry.country)
 
+    @property
+    def temporal_start(self) -> Optional[str]:
```
tillprochaska, line 333: I think I might actually change this (and the respective methods in the TS package) to return a tuple and include the specific property that corresponds to the value returned.

When displaying entities in a timeline, we probably want to display details when users interact with entities, including the label of the start and end property.

followthemoney/schema/Interval.yaml
```diff
@@ -14,6 +14,12 @@ Interval:
     An object which is bounded in time.
   matchable: false
   abstract: true
+  temporalExtent:
+    start:
+      - startDate
+      - date
+    end:
+      - endDate
```
tillprochaska, line 22: Initially, I had planned to use a property as the start *and* end in case of properties that don’t have a corresponding property that can represent the end.

For example, in case of an `Event` with just a `date` (and no `startDate`/`endDate`), it makes sense semantically to assume that it’s a single-day event and thus the `date` is both the start and end.

This doesn’t work well in all cases though, so I decided to only set the temporal extent if a property clearly represents the end of a range (for example `endDate`, `dissolutionDate`).

followthemoney/proxy.py
pudo: I can sort of see how this solves the problem, but my imagination (of the UI code) doesn't yet expand to a place where this problem occurs and isn't fixable otherwise. KISS?

followthemoney/schema/Interval.yaml
pudo: +1 on using convention on the edge case :)

pudo (approved):
I got nothing. This is superb.

followthemoney/proxy.py
tillprochaska: Fair point. The use case for this is displaying both the value and property name of the temporal start for a property:

Screen Shot 2022-11-17 at 15 44 14

It is fixable otherwise, but I think that would mean basically duplicating most of the logic in the Aleph UI, which I’d like to avoid. Any ideas?

pudo: Sold. I vote we merge this whole thing!

PR #867, opened 2022-10-31
Add callfortenders schema (branch)

pudo:


followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,122 @@
...
+      label: "Evaluation mechanism"
+    procurementType:
+      label: "Procurement type"
+    directive:
+      label: "Directive"
+    procedure:
+      label: "Procedure"
+    # What does this mean?
+    # involves:
+    #   label: "Call for tenders involves"
```
mattcg, line 56: This is is a field that's used to differentiate between different types of CfTs. It seems to be a kind of bodge to use the CfT system for other purposes. Here's the label used for it on the PDF version of the Contract Notice: 

image

In this Legal Notice [178 of 2005](https://contracts.gov.mt/en/Resources/Documents/LN178_2005.pdf)  (page 9) you have definitions of DPS and framework agreement.

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,122 @@
...
+    cpvCode:
+      label: "CPV codes"
+      description: "Common Procurement Vocabulary (CPV)"
+      type: identifier
+    # eAuctionsIncluded:
+    #   label: "Inclusion of e-Auctions"
```
mattcg, line 62: Ah, electronic auctions. This is a binary option. If yes then the CfT is a reverse auction. Worth keeping this field for statistical use. You could rename it to "reverse auction included" for example so that it's less vendor specific. Some information on "e-Auctions" in public bidding in Malta: [Reverse Bidding for Data Connectivity across the Public Service- A First for Malta](https://mita.gov.mt/2021/10/05/reverse-bidding-for-data-connectivity-across-the-public-service-a-first-for-malta/)

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,122 @@
...
+    maximumNumberOfLots:
+      label: "Maximum number of lots"
+      type: number
+    euFunding:
+      label: "EU funding"
+    multipleTenders:
+      label: "Multiple tenders will be accepted"
+    # tedLinksForNotices:
+    #   label: "TED links for published notices"
+    #   -> use sourceUrl instead
```
mattcg, line 100: I think this is different to sourceUrl which would be the URL on etenders.gov.mt. TED is the EU tenders journal and I'm not sure every tender that's on etenders (or any other national procurement db) is also automatically on TED. Having this separate field could perhaps allow cross reference to the [TED dataset that's already on Aleph](https://aleph.occrp.org/datasets/755).

followthemoney/schema/CallForTenders.yaml
```diff
@@ -0,0 +1,122 @@
...
+    # Do we need this:
+    fallsUnderGPPScope:
+      label: "Does this call fall under the scope of GPP?"
+      description: "European Green Public Procurement (GPP) or green purchasing."
```
mattcg, line 104: Could be useful. Common to other EU member states.

mattcg (approved):
This is great work. Thank you for the effort involved in making a deep dive into the Maltese CfT system :) I left some comments to answer the questions and doubts.

pudo:
Thanks for the feedback, @mattcg. I've reflected your feedback and would argue this is good to go. Waiting until tomorrow morning for protests :)

PR #871, opened 2022-11-02
Improve domain related email cleanup error handling

### LINKED ISSUE #870: Better error handling regarding invalid domains in e-mail cleanup
I believe https://github.com/alephdata/ingest-file/issues/253 should be fixed here in `followthemoney`, because https://github.com/alephdata/followthemoney/blob/4b102568a3b3d13a8c4b34de4b7dc0f856a08b9c/followthemoney/types/email.py#L71 might [raise a UnicodeError](https://docs.python.org/3/library/stdtypes.html?highlight=encode#str.encode).

The suggested change: return `None` if a `UnicodeError` is raised.

stchris:
This came up in https://github.com/alephdata/ingest-file/issues/253. The `str.encode` function may raise an `UnicodeError` which should be handled. In our case any failure to encode the domain as `idna` means the domain name can't be parsed.

Fixes #870

### DIFF
--- a/followthemoney/types/email.py
+++ b/followthemoney/types/email.py
@@ -68,7 +68,10 @@ class EmailType(PropertyType):
         domain = domain.lower()
         domain = domain.rstrip(".")
         # handle unicode
-        domain = domain.encode("idna").decode("ascii")
+        try:
+            domain = domain.encode("idna").decode("ascii")
+        except UnicodeError:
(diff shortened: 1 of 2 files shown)


pudo (approved):
This definitely feels like the correct place for this!

Rosencrantz (approved):

PR #908, opened 2022-11-21
Schema/Address: Add osm / google places id

simonwoerpel:
I remember discussions about ftm/aleph should NOT start to implement geo functionality, and I am :100: on this side.

But still, as the Address entity has lat/lon props, it would be useful for some applications to be able to add osm or google place id (as a result from a geocoding process)

I don't now if type should be an identifier?


pudo (approved):
Seems simple and useful.

PR #911, opened 2022-11-22
Cache schema.is_a calls

pudo:
There's only about 200 outcomes of this, and it gets called in a bunch of hot loops.


stchris (approved):
Seems good to me. Nothing bad can come from this ™️

PR #913, opened 2022-11-25
Add property type descriptions to JSON model

tillprochaska:
We already include schema descriptions in the JSON model, this improves consistency across types and schemata.

I’ve also converted the property type doc strings from reStructuredText to Markdown. I think the main reason for using reST is the current documentation setup is using Sphinx, but we’re mostly moving away from that and Markdown is a more common format.

### DIFF
--- a/followthemoney/types/common.py
+++ b/followthemoney/types/common.py
@@ -17,6 +17,7 @@ EnumValues = Dict[str, str]
 class PropertyTypeToDict(TypedDict, total=False):
     label: str
     plural: str
+    description: Optional[str]
     group: Optional[str]
     matchable: Optional[bool]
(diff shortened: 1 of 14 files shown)


followthemoney/types/common.py
```diff
@@ -183,6 +184,7 @@ def to_dict(self) -> PropertyTypeToDict:
         data: PropertyTypeToDict = {
             "label": gettext(self.label),
             "plural": gettext(self.plural),
+            "description": gettext(self.docs),
```
tillprochaska, line 187: We currently don’t translate the descriptions and I’m not sure if we ever will, but it doesn’t hurt either…

pudo (approved):

PR #978, opened 2023-01-23
Fix duplicate property label

tillprochaska:
`EconomicActivity` inherits the `Interval:description` property (labeled "Description") and also defines its own `goodsDescription` property (also labeled "Description"). This leads to confusion when displaying lists of properties in user interfaces.

We may want to reconsider if a separate `goodsDescription` property is necessary. However, this would be a breaking change, so for now I suggest we update the label to be unique.


tillprochaska:
Unrelated, but I’ve seen this problem multiple times now and was wondering if we could prevent naming clashes from happening in the future if we added a simple linting script to our CI?

pudo (approved):

PR #1099, opened 2023-05-18
Add ingest-file version and processing timestamp to Document

catileptic:


followthemoney/schema/Document.yaml
```diff
@@ -111,3 +111,11 @@ Document:
...
       label: "Processing error"
       hidden: true
+    processingAgent:
+      label: "Name and version of the processing agent used to process the Document"
+      type: string
+      hidden: true
+    processedAt:
+      label: "Date and time of the most recent ingestion of the Document"
+      type: date
+      matchable: false
```
tillprochaska, line 121: Is there a particular reason all the other `processing*` properties are hidden but this one is not?

PR #1103, opened 2023-05-21
Remove model attribute from Schema and Property

pudo:
This makes it much harder to serialise entities using pickle, so it's nice to get rid of it.


tillprochaska:
@pudo I asked you to hold off releasing this change so we can have a chance to see whether it could cause issues in Aleph. Thanks for being patient, we double-checked and as far as we can see it shouldn’t be a problem :)

pudo:
thanks for checking!

PR #1114, opened 2023-06-01
Display inheritance graph in model explorer

tillprochaska:
I keep referring to the schema YAML files when I need to find child schemata of a specific schema, so in addition to the list of parent schemata, I added a list of child schemata to the explorer pages. Then decided it probably makes sense to display the (partial) inheritance graph as an actual graph.

Screen Shot 2023-06-02 at 00 10 25

**To do:**
- [ ] ~Show siblings~
- [x] Handle schemata without parent or child schemata
- [x] Handle loading state


pudo:
That's hella cool! Wow. Could you even render out a full hierarchy on the landing page for the schema docs?

tillprochaska:
> Could you even render out a full hierarchy on the landing page for the schema docs?

It’s something I’d like to try out, but I’m a bit pessimistic about the result being usable/laid out in a usable way. :D

tillprochaska:
So, I tried to render a full inheritance graph in the same way, and this is the result… I tried to tweak some of the parameters (e.g. to get rectangular edges), but the results are still not usable.

full_graph_test

I implemented this using Mermaid because it’s so simple and I want to introduce it to the Aleph docs too (to be able to easily include flow charts in documentation articles). When I have some more time at some point, I might experiment a bit more, trying to render stuff directly using D3 or a DAG rendering library.

docs/src/components/common/mermaid-theme.css.txt
Rosencrantz, line 1: Why is this css.txt?

Rosencrantz (approved):

PR #1122, opened 2023-06-05
Add de-registration date property for vehicles

mattcg:
Aircraft are sometimes deleted (366 de-registrations from the Malta registry since 2012). Add a field for the de-registration date of a vehicle for when this data point is available.


pudo (approved):
Makes perfect sense!

PR #1126, opened 2023-06-10
RE:  Idea: transform values via python modules in mapping (was: #681)

simonwoerpel:
Hey, I am currently re-considering this: https://github.com/alephdata/followthemoney/pull/681

Still find it very useful. I now added a safe guard via an env var `FTM_SAFE_MODULES` that must contain a comma-separated list of python modules that are allowed (see tests)

So, to use that as an RCE entrypoint, one has to be in control over the followthemoney deployment environment

### DIFF
--- a/followthemoney/mapping/property.py
+++ b/followthemoney/mapping/property.py
@@ -1,15 +1,16 @@
 import re
 from copy import deepcopy
-from normality import stringify
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, cast
-from banal import keys_values, as_bool
 
-from followthemoney.helpers import inline_names
+from banal import as_bool, keys_values
+from normality import stringify
+
 from followthemoney.exc import InvalidMapping
-from followthemoney.proxy import EntityProxy
-from followthemoney.util import get_entity_id, import_method, sanitize_text
-from followthemoney.property import Property
+from followthemoney.helpers import inline_names
 from followthemoney.mapping.source import Record
+from followthemoney.property import Property
+from followthemoney.proxy import EntityProxy
+from followthemoney.util import get_env_list, import_method, sanitize_text
 
 if TYPE_CHECKING:
     from followthemoney.mapping.query import QueryMapping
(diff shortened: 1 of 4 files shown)


pudo:
I really, really hate this idea and I think we should make a conscious decision for FtM not to go down this path ever. Let me explain: I think FtM mappings right now are a really cool idea, because they're a simple no-code way of transforming a (mostly) clean dataset into a set of FtM entities. We also have other, code-based, means to make FtM entities.

This PR destroys that boundary: mappings suddenly depend on non-standard code that doesn't exist in most FtM environments, so they become non-portable immediately. They effectively become a way to tape together custom code for code-based loaders.

The reason for doing this - I assume - is to build more data cleaning functions into mappings. What that fundamentally does is say: "mappings aren't mappings, they're a workflow language for a data cleaning ETL process". But the problem is: the mapping language was never designed to be an ETL workflow language, and it would suck at being one. It doesn't have phases, enough combination abilities etc.

So if we merge this PR it'll end up causing immense pressure to build all the rest of an ETL framework into FtM core. I have studied dozens of ETL frameworks for Python and they're all a mess; there is no reason why one based off of half-arsed feature creep in a mapping language would not be a mess, too. 

Meanwhile, @dchaplinsky started working on an FtM-based ETL framework: https://github.com/dchaplinsky/thebeast - he was never really able to get feedback from the community on that. I would much prefer if we thought about thebeast for a bit and made it ready for purpose, instead of half-assing it into FtM in a way that may cause fatal scope creep to the language. 

Even with module white-listing, I still think it's also a security nightmare.

dchaplinsky:
Hi! While the beast didn't receive enough attention despite all the love and time I put into it, I must admit that it relies on the fqfn imports obtained straight from the yaml file. But since it's modular, you can turn off that part or add a heavy black/whitelisting.

pudo:
Ah but the difference is that the beast isn't attached to a public web API that anyone with an account can write to.

simonwoerpel:
Hey @pudo thanks for the extensive feedback and descussion! I totally understand. I have used this approach here and there and thought it might worth to put "upstream". But your considerations make totally sense. Have to check out the beast, though.

Closing this then :)

PR #1146, opened 2023-06-27
Shorten processingAgent label in Document schema

catileptic:


followthemoney/schema/Document.yaml
```diff
@@ -112,10 +112,11 @@ Document:
       label: "Processing error"
       hidden: true
     processingAgent:
+      label: "Processing agent"
       label: "Name and version of the processing agent used to process the Document"
```
tillprochaska, line 116: Keeping the long version is fine I think, but you’d need to change the key to `description` ;)

followthemoney/schema/Document.yaml
```diff
@@ -112,10 +112,11 @@ Document:
...
       type: string
-      hidden: true
     processedAt:
       label: "Date and time of the most recent ingestion of the Document"
```
tillprochaska, line 119: Maybe shorten that one two (to "Processed at)? Eeven though it’s hidden, we might decide to show it in the future, and it’s also displayed in the FtM model explorer etc.

stchris (approved):

PR #1148, opened 2023-06-28
Maintain insert order of property values

### LINKED ISSUE #1139: Maintain insert order for multi-valued properties
The Python library uses sets to store property values. Python sets do not guarantee order:

```
Python 3.11.3 (main, Apr  7 2023, 21:05:46) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> list({ 3, 2, 1 })
[1, 2, 3]
>>> list({ 1, 2, 3})
[1, 2, 3]
```

This can be a problem in some use cases where the order of values is relevant. For example, ingest-file stores the parts of multipart emails as separate values and the order of the parts is relevant.

This could be resolved by using a different data structure to store property values (e.g. using [dict keys](https://stackoverflow.com/a/53657523))

tillprochaska:
Fix #1139

### DIFF
--- a/followthemoney/proxy.py
+++ b/followthemoney/proxy.py
@@ -86,7 +86,13 @@ class EntityProxy(object):
             if key not in self.schema.properties:
                 continue
             if cleaned:
-                self.add(key, values, cleaned=cleaned)
+                # This does not call `self.add` as it might be called millions of times
+                # in some context and we want to avoid the performance overhead of doing so.
+                seen = set()
+                seen_add = seen.add
+                unique_values = [v for v in values if not (v in seen or seen_add(v))]
+                self._properties[key] = unique_values
+                self._size += sum([len(v) for v in unique_values])
             else:
                 self.add(key, values, quiet=True)
 
--- a/followthemoney/proxy.py
+++ b/followthemoney/proxy.py
@@ -79,18 +79,16 @@ class EntityProxy(object):
         #: than ``id``, ``schema`` or ``properties``, they will be kept in here
         #: and re-added upon serialization.
         self.context = data
-        self._properties: Dict[str, Set[str]] = {}
+        self._properties: Dict[str, List[str]] = {}
         self._size = 0
 
-        for key, value in properties.items():
+        for key, values in properties.items():
             if key not in self.schema.properties:
                 continue
-            if not cleaned:
-                self.add(key, value, cleaned=cleaned, quiet=True)
+            if cleaned:
+                self.add(key, values, cleaned=cleaned)
             else:
-                values = set(value)
-                self._properties[key] = values
-                self._size += sum([len(v) for v in values])
+                self.add(key, values, quiet=True)
 
     def make_id(self, *parts: Any) -> Optional[str]:
         """Generate a (hopefully unique) ID for the given entity, composed
@@ -127,11 +125,10 @@ class EntityProxy(object):
         prop_name = self._prop_name(prop, quiet=quiet)
         if prop_name is None:
             return []
-        return list(self._properties.get(prop_name, []))
+        return self._properties.get(prop_name, [])
 
     def first(self, prop: P, quiet: bool = False) -> Optional[str]:
-        """Get only the first value set for the property, in no particular
-        order.
+        """Get only the first value set for the property.
 
         :param prop: can be given as a name or an instance of
             :class:`~followthemoney.property.Property`.
@@ -217,8 +214,9 @@ class EntityProxy(object):
                     # log.warning(msg, prop.name)
                     return None
             self._size += value_size
-            self._properties.setdefault(prop.name, set())
-            self._properties[prop.name].add(value)
+            self._properties.setdefault(prop.name, list())
+            if value not in self._properties[prop.name]:
+                self._properties[prop.name].append(value)
         return None
 
     def set(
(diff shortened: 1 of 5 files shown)


pudo:
Just so it's on record: this won't live through ftm-store properly, and I'm not sure that can be made to work.

tillprochaska:
> Just so it's on record: this won't live through ftm-store properly, and I'm not sure that can be made to work.

Just to make sure I don’t misunderstand you: I understand that if property values are stored in separate fragments, this PR won’t solve that issue, because order isn’t guaranteed when selecting the individual fragments from the database. That means even after merging this PR, the order of extracted text in `Pages` entities won’t reflect the order in the source document. That’s ok for now and not my main concern.

I’m currently trying to solve a few issues related to previews of multipart emails in Aleph. The plaintext and HTML parts of the email are stored in a single fragment. For example, there would be a single ingest-origin fragment like this:

```json
{
  "schema": "Email",
  "properties": {
    "bodyText": ["First plaintext part", "Second plaintext part"],
    "bodyHtml": ["First HTML part", "second HTML part"],
  },
}
```

As far as I can see, that means that the property values are retrieved in the original order during indexing, then indexed in the original order in ES. Once we retrieve the entity data from ES and instantiate an FtM proxy object in Aleph, the order gets mixed up. That’s what I’m trying to fix.

Does that make sense? (I might still overlook something here, but I think at least for this particular problem, ftm-store shouldn’t be the problem.)

followthemoney/proxy.py
```diff
@@ -79,18 +79,16 @@ def __init__(
...
                 continue
-            if not cleaned:
-                self.add(key, value, cleaned=cleaned, quiet=True)
+            if cleaned:
+                self.add(key, values, cleaned=cleaned)
             else:
-                values = set(value)
-                self._properties[key] = values
-                self._size += sum([len(v) for v in values])
+                self.add(key, values, quiet=True)
```
tillprochaska, line 91: This was previously setting `_properties` directly in case the data is already cleaned. For the sake of simplicity I’m calling the `add` here too. Not sure if this has any disadvantages? (I assume the previous implementation was to avoid unnecessary method calling/loop overhead?)

catileptic (approved):

followthemoney/proxy.py
pudo: Yep this is all super perf optimised. I'd sort of consider keeping it unrolled, this gets called millions and millions of times e.g. when a mapping is generated and saving three lines in exchange for a few hours on the other end may be bad math.

tillprochaska: Yes, makes sense, I’ll change that :)

tillprochaska: I pushed another commit that inlines this again and ran a very basic benchmark based on the `contrib/perf_create_proxy.py` script, creating 2,000,000 proxies with cleaned values. Surprisingly, the inlined version is ~1,4 times slower than the "unoptimized" version (which is only slightly slower than the current FtM release).

Might need to run a slightly more advanced benchmark though to exclude other factors as the reason for this.

### https://github.com/alephdata/followthemoney/pull/1148/commits/f61f7395355b50b83f2cd47f5be471607c4d41a1
Took 10.915786981582642
Took 10.810703992843628
Took 10.882683038711548
Took 10.802056074142456
Took 10.878925085067749

### https://github.com/alephdata/followthemoney/pull/1148/commits/00bb7918169dffa10027a3348ab16d3a7b8eac2e
Took 15.575160026550293
Took 15.75855827331543
Took 15.749438047409058
Took 15.530592918395996
Took 15.721524953842163

### Current followthemoney release
Took 10.602813005447388
Took 10.623345851898193
Took 10.726281881332397
Took 10.77332091331482
Took 10.679194211959839

pudo: Wooo, weird. I guess we technically don't need to `set` but it feels a bit hygienic....

tillprochaska: > Wooo, weird. I guess we technically don't need to set but it feels a bit hygienic....

You mean we wouldn’t need to dedupe the values if `clenaed=True` because we’d expect the consumer to already provide deduplicated values? I think I agree in that the API might not officially guarantee that atm, but it’s been like this forever, so it’s a bit risky to change this now…

tillprochaska: I’m benchmarking a few different scenarios right now, it seems that the results look a bit different when you create a proxy with properties that are already lists for example:

```
{
  "schema": "Person",
  "properties": {
    "name": ["John Doe"],
  },
}
```

```
{
  "schema": "Person",
  "properties": {
    "name": "John Doe",
  }
}
```

tillprochaska: Okay, ran a slightly more sophisticated benchmark. The results suggest that inlining initialization of property values into the constructor indeed has a significant impact. It’s still slower than the currently released implementation, but it’s less worse (only 25% slowdown vs >100% slowdown).

@pudo The numbers below are for 10x creating 1,000,000 proxies. The slowdown sounds tolerable to me, what do you think?

Screen Shot 2023-07-11 at 12 50 57

I’m not an expert in benchmarking/profiling, so this might still be a little flawed. Ran this in Docker containers for ease of testing different Python versions, but can see similar results when running it on the host machine directly.

pudo: Yeah all the non-add ones look totally fine :)

tillprochaska: @pudo Perfect, thanks!

And sorry for bothering you again, but would you mind clarifying what you meant here. I’m not 100% sure I understand what you mean. The set is only used to test whether a value already occurred before, but the values are added to the list of values in their original order.

> Ah but thinking about it a bit more: you can't set there. It's called every time the thing is fished out of the database, which would make this whole PR not persistent....

PR #1167, opened 2023-07-11
Introduce a deprecated flag on props and schemata

pudo:
This creates a mechanism for warning users that a given schema or property will be removed in the future and should no longer be used.


tillprochaska:
It would be great to add this to the `to_dict` methods (so it is included in the model JSON) and ideally to the TS lib, too, so we can display a warning in the model explorer and hide deprecated schemata in the Aleph UI (or at least prevent users from creating new entities using deprecated schemata).

pudo:
I assume this would be really nice to feature in `SchemaPageHeader`, perhaps as an alert box? Maybe I can ask your help for that?

tillprochaska:
Yep, I’ll add it to the docs. Will do that in a separate PR though so this one isn’t blocked. Also, this feature needs to be in a release of the TS lib in order to use it in the docs :)

followthemoney/property.py
```diff
@@ -85,8 +87,11 @@ def __init__(self, schema: "Schema", name: str, data: PropertySpec) -> None:
         self._label = data.get("label", name)
         self._description = data.get("description")
 
+        #: This property is deprecated and should not be used.
+        self.deprecated = as_bool(data.get("deprecated", False))
+
         #: This property should not be shown or mentioned in the user interface.
-        self.hidden = as_bool(data.get("hidden", False))
+        self.hidden = as_bool(data.get("hidden", self.deprecated))
```
tillprochaska, line 94: I’m not sure whether it’s a good idea to make this decision in FtM, maybe it’s better to leave that to the application?

Haven’t thought about how exactly we’d want to handle deprecated schemata/properties in Aleph, but one option would be to display existing entities using the deprecated schema, but disallow users from creating new entities.

followthemoney/schema.py
```diff
@@ -129,9 +132,12 @@ def __init__(self, model: "Model", name: str, data: SchemaSpec) -> None:
         #: inheritance.
         self.abstract = as_bool(data.get("abstract"), False)
 
+        #: This schema is deprecated and should not be used.
+        self.deprecated = as_bool(data.get("deprecated", False))
+
         #: Hide this schema in listings.
         self.hidden = as_bool(data.get("hidden"), False)
-        self.hidden = self.hidden and not self.abstract
+        self.hidden = self.hidden and not self.abstract and not self.deprecated
```
tillprochaska, line 140: This would make any deprecated schema visible, even if it was hidden previously?

followthemoney/property.py
pudo: Removed.

followthemoney/schema.py
pudo: Removed.

pudo:
Agree that linking `hidden` and `deprecated` was sketch :)

tillprochaska (approved):
Nice, thanks!

PR #1170, opened 2023-07-11
Add Position and Occupancy for less repetition of position data

jbothma:
Adds Position and PositionHeld to eventually replace Post.

Need to fix divergent property type and other issues in CI


pudo:
Do we want to put our "active/out-of-office/unsure" classification onto the PositionHeld entity?

jbothma:
> p.s. I proposed https://github.com/alephdata/followthemoney/pull/1167 this morning, which would mark Post as deprecated for a release or two instead of removing it outright. Perhaps you can restore it in this PR so that both discussions are untangled?

Basing on your branch which deprecates Post makes a lot more sense :facepalm: 

Do you mean restore as in rebase? I rebased and pushed. Would it make sense for me to change the PR base branch to yours so my diff is more focused on the additional changes? Or would that just be more confusing since the intention is to merge separately from your PR.

jbothma:
> Do we want to put our "active/out-of-office/unsure" classification onto the PositionHeld entity?

I like that - it's easy to determine this at data entry and remove the need for data users to do the logic. I left unsure out since I think blank covers that - or is there value in an explicit state to say something like "it's not that we didn't bother setting this, but we really didn't know"

jbothma:
> A few fields I could imagine being useful:
>
> Position:inceptionDate
> Position:dissolutionDate
> Position:website or Position:sourceUrl ?
> Position:description ?

Done. I was thinking those dates would be nice too.

jbothma:
Is there a nice obvious way to resolve this?

```
python contrib/check_model.py
DIVERGENT TYPES

  position:
  * Person:position - string
  * PositionHeld:position - entity
```

Is it just a matter of finding another word for it? I don't want to just throw a thesaurus at it if we're creating a jargon jumble that way

jbothma:
I still don't know the reality of querying and summarising data in this schema and what the pros/cons are of an enum type for classifications of positions like

```
national-executive
national-legislature
national-government-administrative
subnational-legislature
party-leadership
diplomatic
```

as opposed to making it a string field which we control on the OpenSanctions crawler side. Also not sure what the migration path looks like if we want to change from our first decision on this.

The main argument I can think of for enum type is validation of values out of the box (right?).

But perhaps that's something we can do easily enough in OpenSanctions code too, and don't have to weigh down FtM users with our special classification decisions?

jbothma:
How do you feel about a number_of_seats property? In wikidata that's on the Cabinet, which for us would be the PublicBody. With a better name it might fit on the Position item in FtM, as in n members of the board of Eskom, n members of parliament of south africa. It would be useful for validation against the number of people considered to actively hold the position.

pudo:
Hey, late session last night :) Some feedback:

a) All FtM props so far at camelCase, let's not start doing snake_case now :) 
b) Maybe PositionHeld:position could be .... drum roll for confusion ... PositionHeld:post ? For light de-escalation we have a lot of entity props that don't match the type name, e.g. Ownership:owner, Directorship:director etc. 
c) Re the `jurisdiction` thing I think we should definitely keep `country` - that's not ambiguous and helpful for rendering the data out. For the sub-unit, I could see `jurisdictionArea`, `territory`, `subnational` all work. 
d) I feel really weird about OccupancyStatus as a single-use-data-type. I get what we're trying to do here but this feels like technical debt in FtM right out the gate. Let me think about this for a bit.

pudo:
One more thing: we could also just introduce new prefixes in `topics`, like `pos.subnational` that work in conjunction with others (`role.pep` + `pos.subnational`).

jbothma:
> One more thing: we could also just introduce new prefixes in topics, like pos.subnational that work in conjunction with others (role.pep + pos.subnational).

Do you mean for high level categorisation?

I like how flexible/powerful combining topics is, but how would we render out useful summaries of positions held if we want something like

- Office of the National Head government
  - President of X
  - Deputy president of X
  - Assistant to the president of X
- Office of Subnational Head of government
  - Governor of Y
  - Governor of Z
  
If that's represented by `gov.exec.head` and `gov.national` or `gov.state` respectively?

jbothma:
Would we want to represent SOE directorship using these schemata **and/or** Directorship?
How about private company directorship? If you see OpenSanctions going down that road for PEPs.

Using Directorship instead of PositionHeld means summarisation of PEP data in OpenSanctions, as well as documentation for data users of how to extract _why_ someone is considered a PEP, and exactly what kind of PEP they are, is a bit more complex, but there might be upsides.

followthemoney/schema/PositionHeld.yaml
```diff
@@ -0,0 +1,32 @@
+PositionHeld:
```
jbothma, line 1: I suspect PositionOccupancy is more correct but it's 7 syllables and I'm a bit worried about our protagonist investigative journalists

jbothma: but it makes the reverse relations a bit clearer:

Position.occupancies
Person.positionsOccupied

followthemoney/schema/Position.yaml
```diff
@@ -0,0 +1,65 @@
...
+  description: >
+    A post, role or position within an organization or body.
+    This describes a position one or more people may occupy
+    and not the occupation of the post by a specific individual at a 
+    specific point in time.
+
+    'subnationalArea' should be used to further restrict the scope of the
+    position. It should not simply represent some regional aspect of the
+    role - e.g. the constituency of a national member of parliament - when
+    their their legislative jurisdiction is nationwide.
```
jbothma, line 15: I'm not sure if I'm overthinking this. My gut tells me it's helpful to decide whether this field is restrictive (the scope of the position is focused on this area) or additive (hey this area is interesting to this position). Happy to delete if it's not helpful now.

pudo: To me, the use case of this is to define subnational levels of government, so I think the description fits extremely well.

pudo (approved):

followthemoney/schema/Position.yaml
```diff
@@ -0,0 +1,65 @@
...
+      type: string
+    description:
+      label: "Description"
+      type: text
+    inceptionDate:
+      label: Inception date
+      type: date
+    dissolutionDate:
+      label: Dissolution date
+      type: date
```
tillprochaska, line 46: It would be great if you could add these to the schema’s temporal extent.

For an example, see https://github.com/alephdata/followthemoney/blob/main/followthemoney/schema/Interval.yaml#L17-L22

followthemoney/schema/Position.yaml
tillprochaska, line 1: It could make sense to extend `Thing` here which already defines most of the common properties like `name`, `description`, `summary`, `wikidataId` etc. But maybe it was an explicit decision not to do that?

followthemoney/schema/Position.yaml
```diff
@@ -0,0 +1,65 @@
...
+      reverse:
+        name: positions
+        label: Positions
+      range: Organization
+    name:
+      label: "Name"
+      type: name
+    category:
+      label: "Category"
+      type: string
```
tillprochaska, line 37: Not sure what the use case for `category` is, but maybe this is already covered by `Thing:keywords`?

tillprochaska:
@jbothma I added a few comments, mostly from Aleph’s perspective (which probably has slightly different requirements compared to OpenSanctions). As Aleph enables end user search across all schemata, we’re trying to reduce the number of properties. Also, Aleph supports graph and time-based visualizations which is why `edge` and `temporalExtent` schema metadata is important to us.

followthemoney/schema/Position.yaml
jbothma: @pudo I was supposed to drop `category` from `Position` to synthesise it for presentation on the summary page, wasn't I?

followthemoney/schema/Position.yaml
jbothma: As discussed in slack, it was mainly because Thing is not normally shown in search results, and not all the Thing properties are relevant. But we have a solution to exclude Positions from search results otherwise, and the other motivations for inheriting from Thing, like data model consistency, make sense.

pudo:
hey @jbothma can you still add the `temporalExtent` and `edge` sections to the `Occupancy` so we can merge this very soon?

jbothma:
> can you still add the temporalExtent and edge sections (as suggested by @tillprochaska) to the Occupancy

@pudo I understood the ask for temporal extent to be on position, and edge on occupancy, which are done.

I assumed Occupancy will inherit temporal extent from Interval because it doesn't define different time properties.

Should I still add temporalExtent to occupancy?

pudo:
No, you're righ!

followthemoney/schema/Occupancy.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - holder
+    - post
+  edge:
+    source: holder
+    label: holds
+    target: post
+    directed: true
+    caption: 
+      - startDate
+      - endDate
```
jbothma, line 21: does anything bad happen if none of the caption fields have a value? I guess if it did, there ought to be a check that at least one of them is required.

tillprochaska: The Python and TS libs will fall back to the schema label if there is no caption property with a value.

So if you have an `Occupancy` with start date `2023-07-21` and end date `2023-07-31`, it would display `2023-07-21` as the edge label. If you have an `Occupancy` without a start or end date, it would display just `Occupancy` as the label.

jbothma:
I'm happy with this if you folks are!

tillprochaska (approved):

PR #1182, opened 2023-07-23
Display warnings for deprecated schemata and properties

tillprochaska:
This shows small warning indicators in schemata and properties tables (and a popup with an explanation on mouseover), and a warning message below the header on schema pages.

**Schemata table:**

Screen Shot 2023-07-23 at 16 09 07

***

**Schema details page:**

Screen Shot 2023-07-23 at 16 08 55

***

**Properties table:**
(I marked `Thing:address` as deprecated only for demonstration because we currently do not have any deprecated properties.)

Screen Shot 2023-07-23 at 16 08 46

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
-        "@alephdata/followthemoney": "^3.2.1",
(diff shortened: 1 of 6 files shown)


pudo (approved):
Looks amazing and the code changes are minimal! :)

PR #1337, opened 2023-12-07
add parent entity reference property to RealEstate

brrttwrks:
RealEstate objects can often have children, such as a building has individual apartments. It often happens that modeling this relationship is important for investigations. This commit and future pull request addess the ability for a RealEstate entity to have a parent so journalists can see subunits of a building, or which apartments/subunits are associated with a single parent unit.


catileptic (approved):
Test pass, the change seems sensible. We could get this merged.

tillprochaska (approved):

PR #1404, opened 2024-02-28
Mapping validation

tillprochaska:


followthemoney/proxy.py
```diff
@@ -209,21 +209,27 @@ def unsafe_add(
         """A version of `add()` to be used only in type-checking code. This accepts
         only a single value, and performs input cleaning on the premise that the
         value is already valid unicode."""
+        original_value = value
+
         if not cleaned and value is not None:
             value = prop.type.clean_text(value, fuzzy=fuzzy, format=format, proxy=self)
+
+        if original_value is not None and value is None:
+            log.warning(f"[{prop.name}] Removed unclean value \"{original_value}\"")
```
tillprochaska, line 218: Note to myself: This is a very simple way to add some logging when unclean property values are removed.

It logs something like this:

```
followthemoney.proxy [WARNING] [phone] Removed unclean value "(567) 992-0000"
```

This isn’t perfect for debugging, it only mentions the FtM property name and the invalid value.

It would be helpful if we could also display the primary key of the entity that is currently being mapped so mapping developers can easily find the relevant source records. Two problems with that though:

1. It would require a few changes to the `PropertyMapping` class as it is currently not aware of the entity mapping it is part of. Also would require changes to `EntityProxy.unsafe_add` to somehow indicate if a value passed cleaning or not.

2. The entity key may not be known at this time, because the key properties haven’t been mapped yet or the cleaned value is itself part of a key property.

tillprochaska: As a low-hanging fruit, we could at least also log the schema name and property type of the affected property:

```
followthemoney.proxy [WARNING] Removed unclean value "(567) 992-0000" from property "Person:phone" of type "phone".
```

tillprochaska:
@brrttwrks @uhhhuh @Okssana This is a first draft of two features based on our recent conversation.

* Optionally disabling property value cleaning in mappings
* Log messages when unclean values are removed

Would appreciate your feedback! :)

followthemoney/proxy.py
tillprochaska: @brrttwrks @uhhhuh @Okssana Any other information that would be helpful when debugging unclean values in mappings?

followthemoney/proxy.py
```diff
@@ -205,26 +205,31 @@ def unsafe_add(
         cleaned: bool = False,
         fuzzy: bool = False,
         format: Optional[str] = None,
-    ) -> None:
+    ) -> Optional[str]:
         """A version of `add()` to be used only in type-checking code. This accepts
         only a single value, and performs input cleaning on the premise that the
-        value is already valid unicode."""
+        value is already valid unicode. Returns the value that has been added."""
```
tillprochaska, line 211: This doesn’t feel like a super clean API decision, but as this method is mostly used internally, that’s probably fine and I couldn’t really come up with a better alternative.

followthemoney/mapping/property.py
```diff
@@ -133,5 +135,18 @@ def map(
...
+                value=value,
+                cleaned=(not self.clean),
+                fuzzy=self.fuzzy,
+                format=self.format,
+            )
+
+            if value is not None and added_value is None:
+                discarded_values.append(value)
+
+        return discarded_values
```
tillprochaska, line 152: Again, I think it’s a little hacky to pass up the discarded values. Alternative might be to pass the `EntityMapping` to the `PropertyMapping` object for context, then do the logging directly in the `PropertyMapping.map` method.

followthemoney/proxy.py
pudo: I don't think it's a bad promise to make... We just want `.unsafe_add` to be super type safe - because all of the rest of the proxy is a bit fuzzy about that.

followthemoney/mapping/entity.py
```diff
@@ -112,16 +115,24 @@ def map(
         # from that accessible to phone and address parsers.
         for prop in self.properties:
             if prop.prop.type == registry.country:
-                prop.map(proxy, record, entities)
+                discarded_values = prop.map(proxy, record, entities)
+                for value in discarded_values:
+                    log.warn(f"[{self.name}] Discarded unclean value \"{value}\" for property \"{prop.prop.qname}\".")
```
pudo, line 120: Would it make sense to add `proxy.id` to these messages? That might make it a bit easier to debug.

followthemoney/mapping/property.py
```diff
@@ -133,5 +135,18 @@ def map(
...
             values = splote
 
+        discarded_values: List[str] = []
+
         for value in values:
-            proxy.unsafe_add(self.prop, value, fuzzy=self.fuzzy, format=self.format)
+            added_value = proxy.unsafe_add(
+                prop=self.prop,
+                value=value,
+                cleaned=(not self.clean),
```
pudo, line 144: This does feel super dangerous to me. It means that no cleaning at all will be done: if the source data says "Russia", that'll be kept (instead of making `ru`), if the date is "Banana", all is considered good. Unless the input data is absolutely spotless, this will generate invalid FtM.

pudo (commented):
I'm a bit concerned about making `clean` disable-able for mappings. Is that really what we're trying to do? It feels like the intent was maybe to introduce a medium level of cleaning ("clean, but not the identifiers") but that isn't what this does - and I also feel like this would be a bit of a code smell....

followthemoney/mapping/entity.py
tillprochaska: Yeah, I originally wanted to do exactly that to make it easier to find the respective row in the source data. Then I realized that some hash of the source data probably isn’t that helpful to a human debugging a mapping…

But… it may still be helpful in case of repeated issues with the same or when using `id_column` which I hadn’t considered before, so I’ll add it.

followthemoney/mapping/property.py
tillprochaska: Yes, you’re probably right. Rather than providing a generic escape hatch for cleaning it might be better to find solutions for concrete issues with validation. I’ll update the PR to drop this specific change, maybe having verbose logging whenever cleaning removes a value is enough for now…

PR #1417, opened 2024-03-26
Hide stub properties in model explorer

tillprochaska:
Stub properties can add a lot of clutter to the properties table, especially for schemata like `Person` or `Company`. This PR hides stub properties by default and provides a toggle to display them when needed.

Screen Shot 2024-03-26 at 10 03 03


pudo:
I love that this is a toggle now. Sometimes you need it, usually not!

PR #1419, opened 2024-03-28
Update links to Aleph documentation

pudo:
It looks like `docs.alephdata.org` is broken for redirects. Using this as an opportunity to update the links to the new docs location.


tillprochaska (approved):
Thanks! Indeed, the redirect got removed by accident. It’s restored now, but updating the links is a good idea nonetheless.

PR #1425, opened 2024-04-02
Improve logging when skipping values or entities in mappings

tillprochaska:
This is a follow-up to #1404. This PR only includes the changes that improve logging as I agree that completely disabling value cleaning can be dangerous.

### DIFF
--- a/followthemoney/mapping/entity.py
+++ b/followthemoney/mapping/entity.py
@@ -129,6 +129,10 @@ class EntityMapping(object):
         # keys.
         proxy.id = self.compute_key(record)
         if proxy.id is None:
+            if self.id_column:
+                log.warn(f"[{self.name}] Skipping entity because no ID could be computed. Make sure that there are no empty values in the \"{self.id_column}\" column.")
+            if self.keys:
+                log.warn(f"[{self.name}] Skipping entity because no ID could be computed. Make sure that there are no empty values in key columns.")
             return None
 
         for prop in self.properties:
@@ -137,6 +141,7 @@ class EntityMapping(object):
                 # the mapping, not in the model. Basically it means: if
                 # this row of source data doesn't have that field, then do
                 # not map it again.
+                log.warn(f"[{self.name}] Skipping entity because required property \"{prop.prop.name}\" is empty.")
                 return None
         return proxy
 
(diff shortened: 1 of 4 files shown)


followthemoney/mapping/entity.py
```diff
@@ -112,16 +115,24 @@ def map(
         # from that accessible to phone and address parsers.
         for prop in self.properties:
             if prop.prop.type == registry.country:
-                prop.map(proxy, record, entities)
+                discarded_values = prop.map(proxy, record, entities)
+                for value in discarded_values:
+                    log.warn(f"[{self.name}] Discarded unclean value \"{value}\" for property \"{prop.prop.qname}\".")
```
tillprochaska, line 120: @pudo You suggested also logging the proxy ID here. I can see how this can be useful, especially when the mapping uses `id_column`. I just don’t know how to implement this in a smart way as the proxy ID hasn’t been computed yet at this point (and it cannot be computed earlier as that requires knowing the property values).

Only option I see is not to log the warning message directly, but cache them until the ID has been computed, something like this:

```python
discarded_values: dict[Property, set[value]] = defaultdict(set)

for prop in self.properties:
  if prop.prop.type == registry:country:
    # Save discarded values
    discarded_values = prop.map(proxy, record, entities)
    discarded_values[prop].update(discarded_values)

for prop in self.properties:
  if prop.prop.type == registry:country:
    # Save discarded values
    discarded_values = prop.map(proxy, record, entities)
    discarded_values[prop].update(discarded_values)

proxy.id = self.compute_key(record)
if proxy.id is None:
  if self.id_column:
    log.warn(...)
  if self.keys:
    log.warn(...)
  return None

if discarded_values:
  # Finally about discarded values here
  for prop, values in discarded_values.items():
    log.warn(...)
```

But that’s a lot of code and indirection for a simple log message… Any good ideas?

pudo: Yeah I think it's too much faff in the core loop of the mapper. Let's leave it.

pudo (approved):
Would it be OK to merge this and make a minor version release?

tillprochaska:
@pudo Thanks for taking another look, just merged the PR! :)

PR #1488, opened 2024-07-04
Always use an array for extends and edge.caption

arp242:
`extends and `edge.caption` was sometimes (not always) a string value when there's just one entry, instead of an array wiht one entry.

This is kind of annoying to parse and deal with; e.g. in Go I need to have a custom type with UnmarshalYAML(). Not hugely complex, but 17 lines of code I could do without. And even in more dynmic languages like Python you'll need an "if" to deal with it.

So just change it to consistently be an array.

### DIFF
--- a/followthemoney/schema/Airplane.yaml
+++ b/followthemoney/schema/Airplane.yaml
@@ -1,7 +1,8 @@
 Airplane:
   label: Airplane
   plural: Airplanes
-  extends: Vehicle
+  extends:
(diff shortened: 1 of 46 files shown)


pudo:
I like this a lot, thanks for the clean-up! Any vetoes on merging this?

pudo (approved):

tillprochaska:
Sounds good!

PR #1497, opened 2024-07-25
Documentation schema: use array for extends

arp242:
Follow-up to #1488; seems I missed this one.


arp242:
Looks like this broke the build because it removed the trailing space; looks like those are also in the .po files and some other places. I can take a look at fixing that later.

PR #1516, opened 2024-08-07
Max typed value length metadata

pudo:
This doesn't actually do anything (i.e. stop `add`), it just creates some benchmark values.

### DIFF
--- a/followthemoney/types/common.py
+++ b/followthemoney/types/common.py
@@ -58,9 +58,10 @@ class PropertyType(object):
     one currently seen by the user."""
 
     max_size: int = 250
-    """The maximum length of a single value of this type. This is used to avoid
-    the individual values that may be malformed or too long to be stored in
-    downstream databases with fixed column lengths."""
+    """The maximum length of a single value of this type. This is used to warn when
+    adding individual values that may be malformed or too long to be stored in
+    downstream databases with fixed column lengths. The unit is unicode codepoints
+    (not bytes), the output of Python len()."""
 
     total_size: Optional[int] = None
     """Some types have overall size limitations in place in order to avoid generating
--- a/followthemoney/types/common.py
+++ b/followthemoney/types/common.py
@@ -196,6 +196,7 @@ class PropertyType(object):
             "label": gettext(self.label),
             "plural": gettext(self.plural),
             "description": gettext(self.docs),
+            "max_size": self.max_size,
         }
         if self.group:
             data["group"] = self.group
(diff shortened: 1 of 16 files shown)


pudo:
We have a bunch of customers who are trying to squeeze our data into SQL tables and so we need to do this enforcement on our end :| 

https://github.com/opensanctions/opensanctions/issues/1045

followthemoney/types/common.py
```diff
@@ -57,7 +57,12 @@ class PropertyType(object):
     other entities that mention the same phone number, email address or name as the
     one currently seen by the user."""
 
-    max_size: Optional[int] = None
+    max_size: int = 250
+    """The maximum length of a single value of this type. This is used to avoid
+    the individual values that may be malformed or too long to be stored in
+    downstream databases with fixed column lengths."""
```
arp242, line 63: Does "length" refer to number of bytes, codepoints, displayed characters, or is it intentionally ambiguous? Would be good to clarify this.

pudo: Fair point! I would angle for codepoints, do you have a preference?

pudo: Postgres calls it characters, perhaps a good reference? https://www.postgresql.org/docs/current/datatype-character.html

arp242: Don't really have a strong preference. I think bytes is probably the most practical as databases with fixed column lengths tend to be in bytes (I think? Not entirely sure) and most RFCs and such tend to specify maximum sizes in bytes. I'm also okay with codepoints.

arp242: Didn't see your second comment – looks like MariaDB also uses "characters" rather than "bytes", so looks like I remembered wrong.

pudo: Added to the comment.

arp242 (approved):

tillprochaska:
@pudo Do you want to (in the future) enforce the max size of individual values in FtM python lib and this is a first step in that direction? Or will the logic that checks that live in OS? (Asking so I know whether I should check typical value sizes in our Aleph instance.)

pudo:
We're playing with adding it to OS as a data quality check: it turns out that when you have a really long name, for example, it's usually a parser error. It's helping us find a ton of bugs. So while I have no personal ambition to push it on you, I feel like it could be a good thing to use in places (e.g. for user-contributed entities, and maybe even in the UI for web-based mappings?)

pudo:
Yeah, absolutely! We're going to probably make it "max length you have to expect from an OpenSanctions dataset" in a few weeks, but that's just one citizen in the ecosystem :)

tillprochaska:
Thanks for the clarification! :)

PR #1517, opened 2024-08-07
Recently introduced field

pudo:
Turns out this should be broader to capture various data sources. I expect this hasn't been deployed to aleph, so it may be a chance to do a quick switch-out.


tillprochaska (approved):
Fine from our end! :)

PR #1518, opened 2024-08-19
Bump identifier length to handle XMR addresses

jbothma:
We'd like to be able to represent the XMR addresses at https://www.treasury.gov/ofac/downloads/sanctions/1.0/sdn_advanced.xml as CryptoWallet:publicKey

They are 96 chars long.

### DIFF
--- a/followthemoney/types/identifier.py
+++ b/followthemoney/types/identifier.py
@@ -26,7 +26,7 @@ class IdentifierType(PropertyType):
     plural = _("Identifiers")
     matchable = True
     pivot = True
-    max_length = 64
+    max_length = 128
 
     def clean_text(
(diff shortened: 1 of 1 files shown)


pudo:
There's a tiny bit of me wondering if we want to have `property.max_length` which defaults to `prop.type.max_length` but can be overridden with more specific details....

pudo:
I think so, if you have a minute. Good examples are probably SWIFT BIC, LEI, OGRN etc!

pudo:
Here's a PR for the more general case: https://github.com/alephdata/followthemoney/pull/1519

jbothma:
That's way nicer, thanks!

PR #1558, opened 2024-10-22
fix: add name to Contract

bamthomas:
Hello,

We are using the aleph FtM models to [generate Java classes](https://github.com/ICIJ/ftm.java) that we are going to integrate into datashare. When the code is generated, there are compilation issues.

I can't see any ways of finding `name` for a `License` object. How could I find the description of this property?

I noticed also that `CallForTenders.title` has no title is it on purpose?

Finally some fields are required but not listed in properties, is it a wanted behaviour?

Thank you for your answers.

### DIFF
--- a/followthemoney/schema/Contract.yaml
+++ b/followthemoney/schema/Contract.yaml
@@ -25,6 +25,9 @@ Contract:
     title:
       label: "Title"
       type: string
+    name:
+      label: "Name"
+      type: name
     authority:
       label: "Contract authority"
       plural: "Contract authorities"


bamthomas:
Ok thank you for your quick answer, for now I had made an approximation that there were not several models inherited at the same time (because Java cannot have multiple inheritance). I have to find another way with interfaces maybe.

bamthomas:
A quick question: for now we are using the required fields to find a parent for Java, to find a unique parent.

It works well because most of the multiple inheritance is ending with diamond pattern, so if we keep a set of inherited attributes, there are no conflicts, and the diamond ends (Document, Thing) with "real" (Class in a sense of Java) entities.

It seems that there is a kind of _underneath_ single inheritance path for each entity. 
Would that make sense to make it explicit?

tillprochaska:
Sorry, I’m not 100% sure about this, maybe @pudo has an answer.

pudo:
Hey @bamthomas this is really exciting work to see, wow. Are you sure you want to 1:1 project the FtM schema onto a Java class hierarchy? In Python we're essentially keeping the values as a map and then wrapping them in some accessor getter/setter logic, which avoids creating massive sparse objects (a given `Person` will have a median of 3-4 props set, and 50 defined) and also allows for legacy or as-yet-unknown props to surf through on the side (e.g. if your version of FtM is older than what the data producer used). 

Regarding the "underneath single path": you mean that there's usually one root, like `Thing` or `Interval`? We do have I think some weirdos that are both (emails? contracts?). It could be possible to add markers for some schema that are essentially `Mixins`, like `Asset` and `Value` - things that don't mean anything, without being part of a bigger whole.

bamthomas:
@pudo nice to see you here and yes we are also excited about FtM models and helping developing a semantic standard for other actors to be able to retrieve easily the data and use it in other tools for example. 

For the 1:1 java class hierarchy it was my first approach because: 
- it was the ©most-simple-thing-that-could-possibly-work (and [it did](https://central.sonatype.com/artifact/org.icij/ftm.java) actually) 
- it was also thinking of [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html)
 we could look at what you did with python, and define a  set of Data Transfer Objects.
 
 It has created a java model like this one (pasted from generated mermaid) 
 
 ```mermaid
 classDiagram
direction BT
class Address {
<<Interface>>

}
class Airplane {
<<Interface>>

}
class Analyzable {
<<Interface>>

}
class Article {
<<Interface>>

}
class Assessment
class Asset {
<<Interface>>

}
class Associate
class Audio {
<<Interface>>

}
class BankAccount {
<<Interface>>

}
class Call {
<<Interface>>

}
class CallForTenders
class Company
class Contract
class ContractAward
class CourtCase
class CourtCaseParty
class CryptoWallet {
<<Interface>>

}
class Debt
class Directorship
class Document
class Documentation
class EconomicActivity {
<<Interface>>

}
class Email {
<<Interface>>

}
class Employment
class Event
class Family
class Folder {
<<Interface>>

}
class HyperText {
<<Interface>>

}
class Identification
class Image {
<<Interface>>

}
class Interest {
<<Interface>>

}
class Interval {
<<Interface>>

}
class LegalEntity
class License
class Membership
class Mention
class Message
class Note {
<<Interface>>

}
class Occupancy
class Organization
class Ownership
class Package {
<<Interface>>

}
class Page {
<<Interface>>

}
class Pages {
<<Interface>>

}
class Passport
class Payment
class Person
class PlainText {
<<Interface>>

}
class Position
class Post
class Project {
<<Interface>>

}
class ProjectParticipant {
<<Interface>>

}
class PublicBody
class RealEstate {
<<Interface>>

}
class Representation
class Sanction
class Security {
<<Interface>>

}
class Similar {
<<Interface>>

}
class Succession
class Table {
<<Interface>>

}
class TaxRoll
class Thing
class Trip
class UnknownLink
class UserAccount
class Value {
<<Interface>>

}
class Vehicle {
<<Interface>>

}
class Vessel
class Video {
<<Interface>>

}
class Workbook {
<<Interface>>

}

Assessment  -->  Thing
Associate  ..>  Interval
CallForTenders  ..>  Interval
CallForTenders  -->  Thing
Company  ..>  Asset
Company  -->  Organization
Contract  ..>  Asset
Contract  -->  Thing
ContractAward  ..>  Interest
ContractAward  ..>  Value
CourtCase  -->  Thing
CourtCaseParty  ..>  Interest
Debt  ..>  Interval
Debt  ..>  Value
Directorship  ..>  Interest
Document  ..>  Analyzable
Document  -->  Thing
Documentation  ..>  Interest
Employment  ..>  Interest
Event  ..>  Analyzable
Event  ..>  Interval
Event  -->  Thing
Family  ..>  Interval
Identification  ..>  Interval
LegalEntity  -->  Thing
License  -->  Contract
Membership  ..>  Interest
Message  -->  Document
Message  ..>  Folder
Message  ..>  HyperText
Message  ..>  Interval
Message  ..>  PlainText
Occupancy  ..>  Interval
Organization  -->  LegalEntity
Ownership  ..>  Interest
Passport  -->  Identification
Payment  ..>  Interval
Payment  ..>  Value
Person  -->  LegalEntity
Position  -->  Thing
Post  ..>  Interest
PublicBody  -->  Organization
Representation  ..>  Interest
Sanction  ..>  Interval
Succession  ..>  Interest
TaxRoll  ..>  Interval
Trip  -->  Event
UnknownLink  ..>  Interest
UserAccount  -->  Thing
Vessel  -->  Thing
Vessel  ..>  Vehicle 
 ```

bamthomas:
So we ended up with defining a `mixin` that helps us to generate code: 

```java
static final Set<String> mixins = new LinkedHashSet<>(List.of("Asset", "Folder", "PlainText", "HyperText"));
```

see [Model.java](https://github.com/ICIJ/ftm.java/blob/main/src/main/java/org/icij/ftm/Model.java).

We'd be happy to talk about this, and will keep you posted on the next steps (this is a beginning with only `required` fields).

PR #1607, opened 2024-12-14
fix: path to TypeScript declarations

sjinks:
In ce9fb7b1438c59badd08bf0493bddafa4aa40d35, `declarationDir` was updated from `./dist/types` to `./dist`. However, `typings` in `package.json` weren't updated.

This PR fixes this issue.


pudo:
Thank you!

PR #1630, opened 2025-01-08
Update astro

tillprochaska:


pudo:
Thank you so so so much for handling this, it's been annoying me a bit but not enough to figure out what needed to be done :) Great to have this in sync again!

tillprochaska:
Yes, sorry it took so long. I’m always a bit torn with mostly static stuff like this. On one hand, there’s no rush to always be on the latest version, on the other hand, I’m annoyed by constant Dependabot notifications for "outdated" development dependencies.

I’ve removed a bunch of customizations from the Astro theem to make it easier to update in the future, and will remove a little more when I have more time.

PR #1699, opened 2025-03-25
Add `proof` to Interval

kdeden:
Limiting `proof` to Thing is problematic:
- It is impossible to express when an Interval is proved by a Document. 
- There's nothing 'Thing'y about a `proof` at all. It's an arbitrary limitation. In the domain of FollowTheMoney investigations an `Ownership` or `Payment` is more likely to be Documented than a `Company` or `RealEstate`.
- As more automatic processes generate entities from unstructured data (ingest-file style), we need to express where these entities came from.

I address this by adding the same property to Interval.

Note: I haven't touched Documentation. The same logic applies but the implementation is more challenging. In Documentation, the range of `entity` should be Thing|Interval, not Thing. That doesn't seem to be possible, as it doesn't seem that Thing and Entity share a common ancestor.

### DIFF
--- a/followthemoney/schema/Interval.yaml
+++ b/followthemoney/schema/Interval.yaml
@@ -72,3 +72,10 @@ Interval:
       label: "Retrieved on"
       type: date
       matchable: false
+    proof:
+      label: Source document
+      reverse:
+        name: proven
+        label: "Derived entities"
+      type: entity
+      range: Document


pudo:
Absolutely love it! There's a world where we would want to make a `Sourced` model that both Interval and Thing derive from, but that feels like a super hefty intervention.....

kdeden:
I guess Thing and Interval are both Entities right? So provenEntities should work. Thanks for prompt response

pudo:
Ah the point was to make the inverses different!

kdeden:
gotcha. also, re the second point about the range of Documentation.entity - I see that you've recently made `extends` into an array. Would we consider doing the same for `range`?

pudo:
That's what I'm trying to avoid. It would require a full rewrite of the graph APIs in aleph and our tooling, and make implementation significantly more complex. Definitely a breaking change - and likely one that would force a bunch of forks. 

Are there other use cases for this?

tillprochaska:
I know FtM isn’t just about Aleph, but from an Aleph user perspective, when we’d make `proven` and `provenIntervals` separate properties, we’d end up with a UX like this in Aleph:

Screenshot 2025-03-27 at 13 30 53

I think most Aleph users probably aren’t aware about specifics of the data model such as the difference between the abstract schemata, so I’m not sure this is the best UX.

That said, if there are reasons to have different inverse property names for other reasons, we shouldn’t compromise just because of that and can workaround the UI issues. (Potentially by simply hiding "Derived intervals" assuming that in most cases a Thing entity related to the Interval would already be listed in "Derived entities" anyway?)

pudo:
I get that there are UX implications - but then why not introduce the meta-type for Thing and Interval? That feels a bit less intrusive than changing the entire data model behind aleph/ftm/yente to be an even more complex hypergraph model... That's not impossible, but it's a really big shift.

tillprochaska:
Ah, sorry, I might have missed something in that case. Would having both `Interval` and `Thing` use the same inverse property name have breaking implications for clients (e.g. in yente)?

I was under the impression that it would in practice work in mostly the same way as having that common meta schema, but if it’s not, then I do fully agree we shouldn’t try to do that!

pudo:
Well, as @kdeden pointed out it makes it impossible to define the `range` of that relationship correctly - it's now two types of schemata. Making `range` an array means that any sort of inverse query needs to be re-written to query multiple ranges going forward.

tillprochaska:
Sorry, that makes total sense. I had thought of that as a separate problem that is only relevant for the `Documentation` question, but it of course also applies to the range(s) of inverse properties.

> There's a world where we would want to make a Sourced model that both Interval and Thing derive from, but that feels like a super hefty intervention.....

Sounds like you’d prefer not to go down that route, did I understand that correctly? I think the UX issue I brought up above is acceptable in that case.

kdeden:
Both raise good points and I think there is general agreement. I'm happy to merge as is and deal with bigger issues as they come up.

To @pudo's question: why not introduce the meta-type for Thing and Interval? It could just be an abstract `Entity` which includes all models, instead of `Sourced` which feels less specific to me, and likely useful in other situations. I can imagine situations where Analyzables and Values and other base models can be 'proofed' or 'Documented'.

tillprochaska:
Trying to summarize the current state of the discussion and options:

**Option 1: Add proof to `Interval`**

* If we add proof to `Interval` directly, we need to chose a different name for the reverse property (e.g. `provenIntervals` as suggested above). (If we were to name the reverse of `Interval:proof` just `proven`, the reverse of `Document:proven` would be either `Thing:proof` or `Interval:proof`, but not both (as expected).

* This is the least invasive option, but it’s a bit ugly because (besides technical reasons) the distinction between proven intervals and proven things feels a bit arbitrary.

* It might be fair to assume that if an `Interval` is proven by a file, it’s likely that there’s also at least one `Thing` entity proven by the same file? For example, if a `Directorship` is proven by a file, there’s like also a `Person` or `Company` entity proven by the same file in which case we could just define the reverse `provenIntervals` as hidden?

**Option 2: Introduce abstract `Entity` schema which is a parent to both `Thing` and `Interval`**

* This options doesn’t require the workaround of naming the reverse props differently.

* Aleph stores a list of an entity’s (parent) schemata in ES for filtering. For example, for a `Person` entity that would include `Person`, `LegalEntity`, `Thing`. If we now add a new parent schema, that list would be `Person`, `LegalEntity`, `Thing`, `Entity` and Aleph instances would have to reindex all entities to ensure this is reflected in ES. However, reindexing may not be possible for large Aleph instances.

* As `Entity` would essentially be a parent schema for any entity, we could work around this by ignoring filters for `schemata:Entity` to avoid the reindex.

**Option 3: Introduce abstract `Sourced` schema which is a parent to both `Thing` and `Interval`**

* Very similar to option 2, but feels more in line with FtM’s current use of abstract "trait" schemata such as `Analyzable`, `Sourced`.

* The same workaround of ignoring `schemata:Sourced` filters is possible in theory, but feels worse given that it’s less clear that `Sourced` is a parent of any other entity (and that might not be true forever).

In summary, there doesn’t seem to be one option that feels 100% right and doesn’t require at least some changes in FtM applications. Wondering if going with Option 1 for now might be the best compromise. It doesn’t require any workarounds in applications and it fulfills the original need (that is being able to link `Interval` entities to a source document).

tillprochaska:
@buffalobuffalobuffalobuffalo Is it ok to move forward with the simple, but not perfectly "clean" solution or do you have concerns about that?

buffalobuffalobuffalobuffalo:
Yes, although I didn't quite grok the bit about "defining the reverse provenIntervals as hidden". In any case, it's not critical to my use-case and I'm in favor of #1.

Thanks both

tillprochaska:
@buffalobuffalobuffalobuffalo I don't have a strong opinion about making the inverse hidden, it was just a quick idea to hide the hacky part of this approach in the UI. I’m also fine with keeping it visible for now, we can still think about a solution if it is actually confusing users.

Can you rename the inverse and then we can merge this?

pudo:
Merged into `develop` with the re-named inverse: https://github.com/alephdata/followthemoney/pull/1749/commits/56c35b00abb46de9f07f427472532a0b0140c65f

PR #1725, opened 2025-04-11
Upgrade to sqlalchemy 2.0 that supports typing

leonhandreke:
Drop dependency on deprecated sqlalchemy2-stubs. This allows type checks
to pass cleanly in zavod (because sqlalchemy2-stubs is no longer
updated).


pudo:
@brrttwrks Is OCCRP still stuck on SQLAlchemy 1.x because of the mess that is `dataset` or have you folks since established a work-around?

PR #1727, opened 2025-04-21
Introduce dataset and catalog metadata handling

pudo:
This is upstreamed from `nomenklatura` in order to make it use by other data producers/consumers. 

Based on the proposed upstream changes for FollowTheMoney 4.0:
* https://aleph.discourse.group/t/proposal-for-followthemoney-4-0/158


pudo:
This is in `develop` via #1728.

PR #1728, opened 2025-04-21
Statement-based data model

pudo:
This brings in the basic idea of a `CompositeEntity` which is made up of `Statements`, each contributing a property value. 

There's some complexity around the notion of `id` statements, which provide checksums of a whole entity in a way that can be used to track changes in the entity. Essentially, a small Merkle tree. 

Output formats CSV and JSON, as well as PACK, which is the internal data format used by OpenSanctions. It's just shoddy CSV. 

TODO: 
* [ ] Explore if the class hierarchy wants to look different between `EntityProxy` and `CompositeEntity`. 
* [x] Bring in some basic CLI commands for statements generation and merging.
* [ ] I've left the SQLAlchemy bindings for statements in NK so far. Maybe they want to move over?

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -7728,9 +7728,9 @@
       }
     },
     "node_modules/vite": {
-      "version": "6.2.5",
-      "resolved": "https://registry.npmjs.org/vite/-/vite-6.2.5.tgz",
-      "integrity": "sha512-j023J/hCAa4pRIUH6J9HemwYfjB5llR2Ps0CWeikOtdR8+pAURAk0DoJC5/mm9kd+UgdnIy7d6HE4EAvlYhPhA==",
+      "version": "6.2.6",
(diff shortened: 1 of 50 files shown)


pudo:
Ok looks like this works pending a `rigour` release to bring in the util functions

simonwoerpel:
Just my 2 cents about SQLAlchemy bindings: I'd see `followthemoney` as the library that defines the model, standards, api interfaces and such. 

The actual sql store implementation could stay/live in NK, and maybe other implementations could live somewhere else (looking at the outdated https://github.com/dataresearchcenter/ftm-columnstore).

pudo:
Agree! The SQL table structure felt too nitty-gritty for FtM. So there's still the question of maybe renaming the `CompositeEntity` to `StatementEntity`. And you raised wanting to import `StreamEntity` from NK, too?

PR #1729, opened 2025-04-21
Adopt PyICU as a dependency

pudo:
I'd like to propose that we make PyICU a dependency of followthemoney. The packaging on the library has improved lots since this project started, and making PyICU optional is a weird source of new developer heisenbugs.


pudo:
Implemented upstream in `rigour` 0.10.1

PR #1749, opened 2025-05-04
Develop (FtM 4.0)

pudo:
Collecting the various PRs here.

### DIFF
--- a/contrib/check_model.py
+++ b/contrib/check_model.py
@@ -4,11 +4,8 @@ from followthemoney import model
 
 
 IGNORE_DIVERGENT_TYPES = [
-    "author",
-    "organization",
     "number",
     "authority",
-    "duration",
     "area",
     "subject",
     "sender",
--- a/followthemoney/schema/CryptoWallet.yaml
+++ b/followthemoney/schema/CryptoWallet.yaml
@@ -31,7 +31,7 @@ CryptoWallet:
       type: date
     currencySymbol:
       label: Currency short code
-    mangingExchange:
+    managingExchange:
       label: Managing exchange
     holder:
       label: Wallet holder
--- a/followthemoney/schema/Sanction.yaml
+++ b/followthemoney/schema/Sanction.yaml
@@ -49,6 +49,7 @@ Sanction:
       label: "Status"
     duration:
       label: "Duration"
+      type: number
     reason:
       label: "Reason"
       type: text


followthemoney/statement/statement.py
```diff
@@ -0,0 +1,237 @@
...
+if TYPE_CHECKING:
+    from followthemoney.statement.entity import SE
+
+
+class StatementDict(TypedDict):
+    id: Optional[str]
+    entity_id: str
+    canonical_id: str
+    prop: str
+    prop_type: str
```
kdeden, line 21: @pudo this is a 1:1 mapping with `prop`, right? So if you know the prop then you know the prop_type? Do we need to keep it?

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,237 @@
...
+    schema: str
+    value: str
+    dataset: str
+    lang: Optional[str]
+    original_value: Optional[str]
+    external: bool
```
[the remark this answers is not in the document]
pudo: The idea of `external` statements comes in with building an [enrichment system](https://www.opensanctions.org/docs/enrichment/). Imagine you have an investigation that you're conducting on a specific domain (place or topic) and you try to expand the entity properties and adjacent entities inside of that investigation with data sourced from much larger, *external*, data sources.

For example: you're doing a country graph on a relevant jurisdiction and have a dozen datasets you include in full (politicians, legal professionals, registered lobbyists, OC figures, etc. etc.), and another set of sources that provide additional context on a subset of these entities: offshoreleaks, company registries, procurement data, cadastral information, etc.

One way you can build an investigative artifact out of this is by materializing a database of the whole datasets, running a dedupe/xref system on it, and then shipping the deduplicated whole graph. That becomes nasty when the enrichment datasets are very large (EGRUL, PSC), or when you don't actually have bulk data access (e.g. OpenCorporates, OCCRP Aleph, Sayari, in practice also Wikidata).

In those cases, it's more effective to run an incremental pull process in which e.g. you query your investigative graph against the given API and store the results as "external", until an analyst has had a chance to verify the match by running a dedupe process on a version of the graph with external statements included. Once verified/integrated, they become an "internal" entity with the next enrichment run, and become part e.g. of the export to a graph DB used for navigating the investigative graph.

followthemoney/compare.py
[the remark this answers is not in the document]
tillprochaska: To summarize: These changes shouldn’t require significant changes in Aleph. That said, it does of course make sense to align the normalization and eventually discontinue use of `fingerprints` in Aleph as well, and if it’s just to reduce legacy code and complexity.

followthemoney/schema/License.yaml
```diff
@@ -19,6 +19,12 @@ License:
   properties:
     area:
       label: "Area"
+      type: number
+      description: "Spatial area of the license or concession"
+    location:
+      label: "Location"
+      type: address
+      description: "Location of the license or concession"
```
tillprochaska, line 27: Is this necessary? `License` already inherits `address` and `addressEntity` from `Thing`.

followthemoney/types/__init__.py
```diff
@@ -22,27 +24,70 @@
...
+    ip = IpType()
+    address = AddressType()
+    date = DateType()
+    phone = PhoneType()
+    country = CountryType()
+    language = LanguageType()
+    mimetype = MimeType()
+    checksum = ChecksumType()
+    identifier = IdentifierType()
+    iban = IbanType()  # TODO: remove
```
tillprochaska, line 45: @pudo I know you wanted to remove the IBAN type before in favor of `identifier` with an IBAN format. Maybe 4.0 is the right time to do that (even though that will require adjustments in Aleph)?

followthemoney/schema/License.yaml
pudo: You're totally right. Let's kick it out.

followthemoney/compare.py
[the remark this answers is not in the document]
pudo: Your concern is totally justified, but has always been an ugly secret of `fingerprints`: it was never a very stable function. Not really meant to be, either, it has too much moving world knowledge inside of it. And it depends on stuff like `pyicu` and `normality` that are also moving. So the fingerprints in the Aleph index was always a crapshoot - a thing that works best if the index is a bit short-lived.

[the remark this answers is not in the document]
tillprochaska: That’s fair enough, but at least I now know the change to `rigour.names` won’t make this issue worse :)

followthemoney/types/__init__.py
tillprochaska: I think ingest-file should be fine, mostly, except that we might want to align the IBAN parsing. Looking at Aleph, indexing/cross-ref, stats/UI are probably the main areas that would need to be adjusted. I could imagine that e.g. having a list of top IBANs is still quite useful, so we might want to compute stats for IBANs even though they aren’t a separate type anymore.

followthemoney/statement/entity.py
```diff
@@ -0,0 +1,407 @@
...
+        raise NotImplementedError()
+
+    @property
+    def caption(self) -> str:
+        """The user-facing label to be used for this entity. This checks a list
+        of properties defined by the schema (caption) and returns the first
+        available value. If no caption is available, return the schema label."""
+        if self._caption is None:
+            self._caption = pick_caption(self)
+        return self._caption
```
tillprochaska, line 150: I’m probably missing something here, but I don’t understand why it’s necessary to override this?

* `StatementEntity` is an `EntityProxy` and inherits the `caption` getter. As `StatementEntity` overrides `get`, shouldn’t that work out of the box?
* The only difference between `pick_caption` and the `caption` getter from `EntityProxy` seems to be that the former caches the result and uses `pick_name` from rigour for name props.

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,237 @@
...
+        "value",
+        "dataset",
+        "lang",
+        "original_value",
+        "external",
+        "first_seen",
+        "last_seen",
+    ]
+
+    def __init__(
```
tillprochaska, line 61: I was wondering whether it would make sense to make this generic and leave some of these attributes to application implementations, in particular things like `first_seen`, `last_seen`, `external`, `lang`?

They might not be relevant for all use cases, and I can imagine that many use cases require additional attributes and thus will have specialized `Statement` classes anyway? For example storing whether a property was provided by an end user/a specific crawler/ingestion stage (unless you model that as separate datasets), application version, user ID, versions for time travel, …

(I do have to admit though that I’m not sure how much complexity this would add given that some of the logic for reading/writing and, in case of `external`, generating IDs, hard-codes these attributes.)

kdeden: Great. cc @pudo 
What would making a Generic look like in usage? You import ftm and then define a generic, then call some method on CompositeEntity to bind it? 
Alternatively, we could define additional statement attributes in a YAML, and use `type()` to construct it with `__slots__`. What do you think?

kdeden: Can you explain a bit more about the problems you've experienced with the complexity of the code? I agree Python's typing system is unwieldy in situations like these, but I also know that the current precedent for addressing this kind of problem is to have `model` reading from YAML and having an `EntityProxy` class to operate on records.

If I understand correctly, you're saying that extending this logic into the structure of the statements themselves is paradigmatically different ('turning FTM into a data modeling language rather than an actual data model'). I think our goal (everyone's goal) is to keep this as a data model, but make sure it's extensible, and not tell people what they ought to do with it. 

On that note, FtM consumers will want to store other information that is related to their use-case on a statement level, (e.g. as you have with `external` and `canonical_id`). I think that's in line with the general vision, and I'm mainly interested in how to implement this technically.

re desigining against an unknown set of requirements -- OCCRP has a good idea of what we'd want out of a statements-as-rows database model (cc @gondzo, but it's not something we'd want to enforce on others. If it's technically feasible, I prefer to decide on a *ballpark* domain of database logic FTM should support and find a way to leave that mostly open.

If it's not technically feasible, the other option would be to pack the Statement model with all of the potentially  attributes we can think of now, and continue updating as the need arises from the community. (e.g. `is_ai_generated`, `version`, `fact_checked`).

Anyway, can you share a bit more about those limitations you encountered? How did you get to generics of generics?

pudo: The thing that I felt about generic statements was that they become very concrete in every single place you use them: the [serialization/deserialization code](https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/statement/serialize.py#L20-L34) needs to be different, any sort of [database storage](https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/statement/db.py#L24-L36) is different, the way you'd [expose them through the statement proxy](https://github.com/alephdata/followthemoney/blob/develop/followthemoney/statement/entity.py#L219-L220) is different. The place where I ran into real type trouble was around trying to describe stuff like an abstract [graph data backend](https://github.com/opensanctions/nomenklatura/blob/main/nomenklatura/store/base.py#L12) (for context: `CE` in that code is `SE` in this PR), which then becomes generic on like two levels (`CE` is generic over `S`). Then you get to do all this in [multiple languages](https://github.com/alephdata/followthemoney/blob/develop/java/src/main/java/tech/followthemoney/statement/Statement.java#L17-L28).

I'm aware that I'm coming off super aggressive/defensive in discussing the statements stuff, but part of the reason for that is the realization (painfully gathered over the last four years) that making statement data work in prod is a hellish bit of work. We've churned through Postgres, BigQuery, Redis, DuckDB as backends, and nothing works well. That's sort of what makes me think that in order to get this to work, we should either make one structure that is super opinionated, or - hear me out - two: one that's basic in descriptiveness but works for large datasets (like [EGRUL](https://data.opensanctions.org/artifacts/ru_egrul/20250512190833-gym/statements.pack)), and one that has full "properties for properties" support ([wikidata style](https://www.mediawiki.org/wiki/Wikibase/DataModel/Primer#Statements)), but does not have the goal of processing bulk data (instead focus on smaller datasets). Wikidata is a good case study of this: loading scientific citation data basically broke their database in terms of scale, so now they're trying to avoid bulk data imports. That same data, [when formatted as FtM](https://followthegrant.org/data) is basically mid-size for the sort of material we're sometimes dealing with.

tillprochaska: > The broader point I'd make is: part of the objective here is to evolve FtM as an interchange format as much as an internal domain model. Making the core element of it have a "proprietary specials" section sort of abandons that objective a bit. Do we want that?

Thanks for clarifying and for sharing your experience regarding the complexity of a generic implementation, that makes sense. Also agree that the statement model here doesn’t need to map 1:1 to internal storage, and this resolves some of the use cases I mentioned above. For example, the ID of the user who added a property value can simply be dropped when serializing to statements. This probably also applies to stuff like the version of FtM or ingest pipeline used which isn’t useful beyond system boundaries.

There are other use cases where that decision isn’t 100% obvious, for example: In Aleph’s current fragment model, each fragment has an `origin` which can be used to designate that the fragment originated from a certain stage of the ingest pipeline, a particular scraper, an ML/AI model, a user, … This might be useful to keep when sharing statements, but maybe it’s also fine to just drop that information.

Then there are some attributes for specific use cases (such as `fact_checked` as mentioned @kdeden above. @kdeden What do you think about the third option of treating these as internal implementation details for now, then potentially upstreaming them later?

followthemoney/statement/entity.py
tillprochaska: So to state my question explicitly: Is there a reason why we can’t have the `caption` getters share the same implementation (maybe except for the caching part if sensible).

followthemoney/types/__init__.py
tillprochaska: > FWIW there's a concept of "strong identifiers" no

That’s only in rigour for now, right?

followthemoney/statement/entity.py
pudo: No. I think medium-term if may be interesting to `pick_caption` to take tuples of `(name, lang?)` so that we can pre-select english names when that metadata is available, but that can still be one implementation.

pudo:
OK, I've just adopted most of @tillprochaska's feedback:

* Remove `License:location`
* Remove `registry.iban` 
* Remove `StatementEntity.caption` (or rather: move it upstream)
* Remove `Statement.prop_type` (it's there as a property, and in the CSV export format)

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,239 @@
...
+            first_seen=datetime_iso(row.first_seen),
+            external=row.external,
+            last_seen=datetime_iso(row.last_seen),
+        )
+
+    @classmethod
+    def from_entity(
+        cls,
+        entity: "StatementEntity",
+        dataset: str,
```
kdeden, line 212: I notice that this works perfectly well with the old fashioned `EntityProxy`. Is there an issue with expanding this signature to include that? Seems like a great compatibility feature.
```suggestion
        entity: "EntityProxy",
```

kdeden: Example:

```
person: EntityProxy = model.make_entity("Person")
person.make_id("test_person")
person.add("name", "John Doe")
person.add("birthDate", "1980-01-01")
person.add("email", "john.doe@example.com")

for statement in Statement.from_entity(person, dataset="test"):
    print(statement.to_dict())

=====OUTPUT=====
{'canonical_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'entity_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'prop': 'id', 'schema': 'Person', 'value': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'dataset': 'test', 'lang': None, 'original_value': None, 'first_seen': None, 'last_seen': None, 'external': False, 'id': '9385d7840a5a553d58a5a5e10491b9d56cdc35e8'}
{'canonical_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'entity_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'prop': 'name', 'schema': 'Person', 'value': 'John Doe', 'dataset': 'test', 'lang': None, 'original_value': None, 'first_seen': None, 'last_seen': None, 'external': False, 'id': 'ab5a7ceab0ea9342546f2acaa18d9e9605328303'}
{'canonical_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'entity_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'prop': 'birthDate', 'schema': 'Person', 'value': '1980-01-01', 'dataset': 'test', 'lang': None, 'original_value': None, 'first_seen': None, 'last_seen': None, 'external': False, 'id': 'c7afdd693fe54b30ae043a962b84ba1e10b94fe9'}
{'canonical_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'entity_id': 'f73e8656f8a7a531b33d474fc54f9759e3343bf5', 'prop': 'email', 'schema': 'Person', 'value': 'john.doe@example.com', 'dataset': 'test', 'lang': None, 'original_value': None, 'first_seen': None, 'last_seen': None, 'external': False, 'id': '66165c4fbf27beb339b72141669cce7a4edddede'}
```
No data is added, but the format is compatible.

followthemoney/statement/entity.py
```diff
@@ -0,0 +1,407 @@
...
+                self.add(prop, values, cleaned=True, quiet=True)
+            return self
+        for stmt in other._iter_stmt():
+            if self.id is not None:
+                stmt.canonical_id = self.id
+            self.add_statement(stmt)
+        self.extra_referents.update(other.extra_referents)
+        return self
+
+    def to_dict(self) -> Dict[str, Any]:
```
kdeden, line 353: FTM producers are going to need this, right @pudo? I also have your ftm 'data-lake' spec in mind here too.

```suggestion
    def to_statements_dict(self) -> Dict[str, Any]:
        data: Dict[str, Any] = {
            "id": self.id,
            "schema": self.schema.name,
            "statements": [s.to_dict() for s in self._iter_stmt()],
        }
        return data
```

followthemoney/statement/statement.py
kdeden: Actually, it doesn't seem to work in the expected way for the `StatementsEntity`, because the members of `Statement` are set without regard for the input values.

e.g.
```
    person: StatementEntity = StatementEntity(
        Dataset.make({"name": "testdataset"}),
        {"schema": "Person", "properties": {}},
    )
    person.id = "test_id"
    person.add_statement(
        Statement(
            entity_id="test_id",
            prop="education",
            schema="Person",
            value="University of California, Los Angeles",
            dataset="testdataset",
            first_seen="2025-01-01",
            last_seen="2025-01-01",
        ),
    )
    person.add_statement(
        Statement(
            entity_id="test_id",
            prop="education",
            schema="Person",
            value="University of California, Los Angeles",
            dataset="testdataset",
            last_seen="2025-01-02",
        ),
    )
    print("--------------------------------")
    for s in Statement.from_entity(person, dataset="testdataset"):
        print(s.to_db_row())
```
returns
```
{
    'canonical_id': 'test_id',
    'entity_id': 'test_id',
    'prop': 'id',
    'schema': 'Person',
    'value': 'test_id',
    'dataset': 'testdataset',
    'lang': None,
    'original_value': None,
    'first_seen': None,
    'last_seen': None,
    'external': False,
    'id': '54cbed65f3a51c2c859014bebd91529f0a77557b'
}
{
    'canonical_id': 'test_id',
    'entity_id': 'test_id',
    'prop': 'education',
    'schema': 'Person',
    'value': 'University of California, Los Angeles',
    'dataset': 'testdataset',
    'lang': None,
    'original_value': None,
    'first_seen': None,
    'last_seen': None,
    'external': False,
    'id': '2509f49fd78932551c199f1b9f47a91aafa40b22'
}
```
So none of the data from the Statements other than the value has propagated. Am I missing something here @pudo?

followthemoney/statement/entity.py
```diff
@@ -0,0 +1,407 @@
...
+            values,
+            cleaned=cleaned,
+            quiet=quiet,
+            fuzzy=fuzzy,
+            format=format,
+            lang=lang,
+            original_value=original_value,
+        )
+
+    def add(
```
kdeden, line 201: In my conversations with data journos and people producing ftm data, we would love a class that we can use as an EntityProxy but with Statements, and use it in the same way to produce FTM data:
```
entity = model.make_statement_entity("Person")
entity.make_id("test_person")
entity.dataset = "testdataset"
entity.add("name", value="John Doe", last_seen="2025-01-01", origin="some process")
entity.add(
    "education",
    value="Universidad de Santiago de Chile",
    last_seen="2025-01-01",
    lang="es",
    # no need to specify the dataset/schema/entity_id, it will be added automatically
)
```
Could that be this class? It seems like this class is only meant to be used by FTM consumers. Is that intentional?

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,239 @@
...
+from typing import cast, TYPE_CHECKING
+from typing import Any, Dict, Generator, Optional
+from typing_extensions import TypedDict, Self
+from rigour.time import datetime_iso, iso_datetime
+from rigour.boolean import bool_text
+
+from followthemoney.statement.util import get_prop_type, BASE_ID
+
+if TYPE_CHECKING:
+    from followthemoney.statement.entity import StatementEntity
```
kdeden, line 13:
```suggestion
    from followthemoney.proxy import EntityProxy
```

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,239 @@
...
+        "prop",
+        "schema",
+        "value",
+        "dataset",
+        "lang",
+        "original_value",
+        "external",
+        "first_seen",
+        "last_seen",
+    ]
```
kdeden, line 57:
```suggestion
        "last_seen",
        "origin",
        "fragment",
        "extra",
```

followthemoney/statement/statement.py
```diff
@@ -0,0 +1,239 @@
...
+        dataset: str,
+        lang: Optional[str] = None,
+        original_value: Optional[str] = None,
+        first_seen: Optional[str] = None,
+        external: bool = False,
+        id: Optional[str] = None,
+        canonical_id: Optional[str] = None,
+        last_seen: Optional[str] = None,
+    ):
+        self.entity_id = entity_id
```
kdeden, line 74:
```suggestion
        origin: Optional[str] = None,
        fragment: Optional[str] = None,
        extra: Optional[Dict[str, str | int | float | bool | None]] = None,
    ):
        if extra is not None:
            for key in extra.keys():
                if key in self.__slots__:
                    raise ValueError(f"Key {key} is a reserved slot")
        self.extra = extra
        self.origin = origin
        self.fragment = fragment
```

catileptic:
@kdeden I'm looking at refactoring OpenAleph to use FTM 4.0. I was wondering how much work you foresee this PR will still take? A very rough estimate is absolutely fine :)

kdeden:
Thanks for the nudge @catileptic . I anticipate it done this week, by Friday. It's really a matter of 
a) adding all the convenience methods and constructors and making sure this new `Statement`, `StatementDict` and `StatementEntity` function as expected.
b) maybe adding another test for serializing and deserializing statements
Maybe @pudo and I can get through it in one session tomorrow?

followthemoney/statement/entity.py
pudo: Implemented this with some tests in https://github.com/alephdata/followthemoney/pull/1749/commits/b1a1b4c6e48eb5b8f6294e5c1393829dbcd52c58

followthemoney/statement/statement.py
pudo: Not to be a pest, but: what's the meaning of fragment, here? To me, the charme of statements is that they get rid of fragments (both mechanisms do the same thing, but statements do it a bit more elegantly).

followthemoney/statement/statement.py
kdeden: @pudo 
I can modify this to use EntityProxy instead, which basically 'upgrades' an FTM <4 statement to 4.0, is that the intended usecase? 
Or it can use StatementEntity and in that case, I feel like the statement data should propagate down, unless I'm missing something? Just lmk

kdeden:
* I rebased this branch to incorporate the latest changes from the main branch. If you have anything last minute to add (unlikely) you may need to resync. 
* The docs build failing is due to hyperlink finding broken links, although the broken links are all internal and directory structure hasn't changed. I will investigate this again in a few hours.

PR #1753, opened 2025-05-09
Upgrade python-stdnum

jbothma:


pudo:
FWIW kicked it from dependencies entirely

PR #1755, opened 2025-05-13
Numbers with units parsing

pudo:
This is an attempt to implement what I understand to be @tillprochaska's current approach to number parsing in FtM. It's not being adopted in `clean_text` for the moment, to avoid the shock of suddenly making vast sections of mapped data invalid. Should we make it required?

### DIFF
--- a/followthemoney/types/number.py
+++ b/followthemoney/types/number.py
@@ -13,7 +13,17 @@ class NumberType(PropertyType):
     In the future we might want to enable annotations for format, units, or
     even to introduce a separate property type for monetary values."""
 
-    CAST_RE = re.compile(r"[^0-9\-\.]")
+    DECIMAL = "."
+    SEPARATOR = ","
+    PRECISION = 2
+
+    _NUM_UNIT_RE = (
+        f"(\\s?\\-?\\s?\\d+(?:{re.escape(DECIMAL)}\\d+)?)\\s*([^\\s\\d][^\\s]*)?"
+    )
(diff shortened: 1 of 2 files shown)


tillprochaska:
I’ll take a look at the beginning of next week!

followthemoney/types/number.py
```diff
@@ -22,9 +32,55 @@ class NumberType(PropertyType):
...
+        match = self.NUM_UNIT_RE.match(value)
+        if not match:
+            return None, None
+        number, unit = match.groups()
+        if unit is not None:
+            unit = unit.strip()
+            if len(unit) == 0:
+                unit = None
+        # TODO: We could have a lookup table for common units, e.g. kg, m, etc. to
+        # convert them to a standard form.
```
tillprochaska, line 54: As mentioned before, I think not checking whether a unit is actually valid for a given property is good enough, but I’m also not opposed to that idea, so if you think it would be useful/necessary, that’s absolutely fine from my perspective!

tillprochaska (approved):
First of all, sorry for replying later than promised, and thanks for making the first step. This looks good to me!

> It's not being adopted in clean_text for the moment, to avoid the shock of suddenly making vast sections of mapped data invalid. Should we make it required?

That’s a good point. Given that this targets the 4.0 branch which will include some other breaking changes, too, I think it would be acceptable to make it required. Realistically, going forward FtM applications will have to handle invalid values in some way (e.g. by logging/warning, storing invalid values separately, …).

Although there’s of course no rush here, and we can always make it required in a later version.

PR #1772, opened 2025-06-07
feat: ensure valid dataset names

simonwoerpel:
This ensures dataset names are valid. This means:
- only lowercase ascii characters
- no whitespace
- no other special chars than `_`
- no `_` at the beginning or end
- (see test)


pudo:
Awesome, this is really helpful!

